PR #6, opened 2021-04-08
PR for Feedback

timomitchel:


app/facades/results_facade.rb
```diff
@@ -0,0 +1,12 @@
+class ResultsFacade
+  def self.results(q)
+    @foods = FoodService.new.results(q).first(10).map do |data|
+        Food.new(data)
+      end
```
timomitchel, line 5: This is great. We want to take the first 10 from the results and then create our Food poros with that data. No need for the `@foods` instance variable though. Instance variables are only passed from controllers to views. The return of your `map` which is the array of Food objects is returned to your controller action that calls this method.

app/services/food_service.rb
```diff
@@ -0,0 +1,23 @@
+class FoodService
+  def results(q)
+    response = conn.get("/fdc/v1/foods/search?") do |faraday|
+      faraday.params['query'] = q
```
timomitchel, line 4: Love that you're setting your query params here in the block. Readable and maintainable.

app/services/food_service.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+    parsed = JSON.parse(response.body, symbolize_names: true)[:foods]
```
timomitchel, line 6: This could parser could be dried up if we put it in its own method.

app/services/food_service.rb
```diff
@@ -0,0 +1,23 @@
...
+      faraday.params['query'] = q
+    end
+    JSON.parse(response.body, symbolize_names: true)[:totalHits]
+  end
+
+  private
+  def conn
+    conn = Faraday.new(url: "https://api.nal.usda.gov") do |faraday|
+      faraday.params["api_key"] = ENV['usda_foods_api']
+    end
```
timomitchel, line 21: Great job setting the connection object up with your API key here.

app/services/food_service.rb
```diff
@@ -0,0 +1,23 @@
+class FoodService
+  def results(q)
```
timomitchel, line 2: This method and the total_results method would be better off as class methods, so that Ruby has less garbage collection to do. https://scoutapm.com/blog/ruby-garbage-collection

app/services/food_service.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+    parsed = JSON.parse(response.body, symbolize_names: true)[:foods]
+
+  end
+
+  def total_results(q)
+    response = conn.get("/fdc/v1/foods/search?") do |faraday|
+      faraday.params['query'] = q
+    end
+    JSON.parse(response.body, symbolize_names: true)[:totalHits]
```
timomitchel, line 14: If the search term is the same, which I assume that it is based on the spec, then we actually don't need to hit the API twice with this method and the. We should have one service method that returns the raw JSON to the API. This way, we only ever make one request. Then, our facade can dig into both `:foods` and `:total_hits`, and we could return that to the controller as a 2D array. Something like:
`[30,000, [FoodObject1, FoodObject2,...]]` 
Then, we can deconstruct this array with one line  of syntactic sugar in our controller:
`@total_hits, @foods = FoodFacade.results`
The entire benefit being that we only make one API call, and this speeds up our application.

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,14 @@
...
+<ul>
+  <section class='results'>
+    <% @foods.each do |food| %>
+    <li>GTIN/UPC code: <%= food.gtin %></li>
+    <li>Description: <%= food.description %></li>
+    <li>Brand Owner: <%= food.brandowner %></li>
+    <li>Ingredients: <%= food.ingredients %></li>
+    <hr>
+    <% end   %>
+  </section>
```
timomitchel, line 13: This view looks great.

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+      fill_in :q, with: 'sweet potatoes'
+      click_button 'Search'
+      expect(current_path).to eq('/foods')
+      expect(page).to have_content('Total search results: 40788')
+      within(first(".results")) do
+        expect(page).to have_content("GTIN/UPC code: 492111402857")
+        expect(page).to have_content("Description: SWEET POTATOES")
+        expect(page).to have_content("Brand Owner: ARCHER FARMS")
+        expect(page).to have_content("Ingredients: SWEET POTATOES.")
+      end
```
timomitchel, line 18: Clean, and effective testing here.

spec/model/food_service_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+      expect(results.size).to eq(50)
+      expect(results.class).to eq(Array)
+    end
+
+    it '#total_results' do
+      results = FoodService.new.total_results('sweet potato')
+
+      expect(results).to eq(40788)
+    end
```
timomitchel, line 19: Again, clear and efficient testing.

spec/model/food_service_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+end
```
timomitchel, line 21: Would have loved to see a Facade, and PORO test as well.

PR #9, opened 2021-05-24
George Ho - mid mod

georgehwho:
Add:
- food index page which takes food search and spits out info
- food service
- food facade
- food list poro
- food item poro
- gems: webmock, VCR, and json
- tons of tests 📈

### DIFF
--- a/app/poros/food_list.rb
+++ b/app/poros/food_list.rb
@@ -1,0 +1,9 @@
+class FoodList
+  attr_reader :total_hits,
+              :foods
+
+  def initialize(food_hash)
+    @total_hits = food_hash[:totalHits]
+    @foods = food_hash[:foods].first(10).map { |food| FoodItem.new(food) }
+  end
+end
--- a/spec/services/food_service_spec.rb
+++ b/spec/services/food_service_spec.rb
@@ -1,0 +1,17 @@
+require 'rails_helper'
+
+describe 'food service test' do
+  it 'can make a call to the api', :vcr do
+    json_response = FoodService.request_api("search?query=sweet potatoes")
+
+    expect(json_response).to be_a Hash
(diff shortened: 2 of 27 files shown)


app/poros/food_list.rb
```diff
@@ -0,0 +1,9 @@
+class FoodList
+  attr_reader :total_hits,
+              :foods
+
+  def initialize(food_hash)
+    @total_hits = food_hash[:totalHits]
+    @foods = food_hash[:foods].first(10).map { |food| FoodItem.new(food) }
+  end
+end
```
megstang, line 9: I really like this idea, George. Because ideally, we're just sending one object to the view, and you're combining everything we need for the view right here in this one object. You could also consider using a parameter in your API call to limit how many foods you get back for that first page, but I'm also okay with this, as you're limiting it to the first ten before iterating and creating the FoodItem. This is a chunk of logic in your initialize method though, so I'd also suggest putting this line 7 into a method instead.

spec/services/food_service_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+    expect(json_response).to be_a Hash
+    expect(json_response[:foodSearchCriteria][:query]).to eq('sweet potatoes')
+  end
+
+  it 'can search for sweet potatoes', :vcr do
+    json_response = FoodService.search("sweet potatoes")
+
+    expect(json_response).to be_a Hash
+    expect(json_response[:foodSearchCriteria][:query]).to eq('sweet potatoes')
+  end
```
megstang, line 16: You can also test in here for the key's/values that you'd expect to get returned for your use later.

megstang (commented):
@georgehwho Great job on this, I left just a few comments, so please let me know if you have any questions. I'll follow up over slack regarding scoring.

PR #10, opened 2021-05-24
Wyatt Wicks - Refactoring

Wyattwicks:
This is my re-factored code, I ran out of time to fix what was happening in my views, and got sloppy with TDD. 
It wont let me also pull request my main branch, but I looked in the commits and what I had originally can be seen in those, up to Add Working API Call. Thanks!


app/poros/food_index.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def initialize(data)
+    @totalhits = data[:totalHits]
+    
+    @food_info = data[:foods].first(10).map do |food_info|
+      @description = food_info[:description]
+      @gtinupc = food_info[:gtinUpc] 
+      brand = food_info[:brandOwner]
+      ingredients = food_info[:ingredients]
+    end
```
timomitchel, line 14: The iteration belongs in the facade, and `food_info` would be passed to each `Food` object where it would be responsible for storing its own state.

app/facades/food_facade.rb
```diff
@@ -0,0 +1,9 @@
+class FoodFacade
+  class << self 
+    def search_foods(food)
+
+      data = FoodService.call_for_a_food(food)
+      FoodIndex.new(data)
```
timomitchel, line 6: Would have been a good idea to extract the total_hits from the instantiation of your `Food` objects here. `FoodIndex` is not SRP.

timomitchel: One thing to consider would be to return a 2d array from the facade where the structure looks like this: 
`[<num_of_total_hits>, [FoodObject1, FoodObject2...]]`
Then you could unpack this in your controller like so: 
`@total_hits, @top_10_foods = FoodFacade.search_foods(food)`

app/services/food_service.rb
```diff
@@ -0,0 +1,17 @@
+class FoodService
+  class << self
+    def call_for_a_food(food)
+      response = conn.get("/fdc/v1/foods/search?&query=#{food}&api_key=#{ENV['food_api_key']}")
```
timomitchel, line 4: Would be better to set the params in a block here. Something like:
```
conn.get("/fdc/v1/foods/search") do |req|
   req.query = food
   req.api_key = ENV['food_api_key']
end
```

Much more readable and maintainable like this.

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,6 @@
+<h1>Results: <%= @foods.totalhits %> </h1>
+
+<%= @foods.food_info.each do |food| %>
+<%= @description %>
+<%= @gtinupc %>
```
timomitchel, line 5: These instance variables would only be passed to the view if they were set in the controller. You don't have access to instance variables you set in your `FoodIndex` without an attr_reader, and then you would use the method dot syntax.

coverage/.last_run.json
```diff
@@ -0,0 +1,5 @@
+{
```
timomitchel, line 1: Best to add everything in `/coverage` to the `.gitignore`

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+RSpec.describe "Foods page" do
+  
+  it "shows all foods that match a search" do
+    visit root_path
+    fill_in :q, with: "Sweet Potatoes"
+    click_button 'Search'
+    expect(current_path).to eq(foods_path)
+    expect(page).to have_content('Results:')
+    expect(page).to have_content(44128)
+    expect(page).to have_content('SWEET POTATO')
```
timomitchel, line 11: Definitely needed a bit more in terms of testing all  the attributes and the number of foods (10) the spec was asking for.

spec/services/food_service_spec.rb
```diff
@@ -0,0 +1,10 @@
+RSpec.describe 'Food Services' do
+  describe 'food_search(food)' do
+    it 'returns relevant foods' do
+      search = FoodService.call_for_a_food("sweet potatoes")
+      expect(search).to be_a(Hash)
+      expect(search[:results]).to be_an(Array)
```
timomitchel, line 6: This is on the right track, but we would also want to dig into the `search[:results]` array a bit deeper for all of the info that we plan on storing in our Food objects as well as the data type of `total_results`.

timomitchel (commented):
This is on its way but you ran out of time. Think about taking what you know about SRP and using your tests to guide your implementation. I recommend finishing the refactor, and adding test to complete test both functionality and each class you created as part of your refactoring pattern.

PR #11, opened 2021-05-24
Aidan McGuire

ghost:
This didn't go very well... I don't think I understood the prompt until it was too late.


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,5 @@
+class FoodsController < ApplicationController
+  def index
+    @foods = FoodsService.foods_search(params[:q])
```
megstang, line 3: Looks like you may have run out of time to create a facade, but ideally, we'd call the Facade which would then call our Service. We can think of the Service as being responsible for making the call to gather the data we need, and then our facade formatting it, creating objects out of it if needed, in order to send one object to the view with only the data we need to show.

app/services/foods_service.rb
```diff
@@ -0,0 +1,24 @@
+class FoodsService 
+  class << self
+
+    def foods_search(food)
+      response = conn.get('/fdc/v1/foods/search') do |req|
+        req.params['query'] = "ingredients:#{food}"
+        req.params['pageSize'] = 10
+      end
```
megstang, line 8: I like how you're adding the pageSize into your query here since, for now, we know we only want to display 10 foods from this search.

app/controllers/foods_controller.rb
megstang: Currently, since you're not making objects, and extracting the data you want from the parsed JSON, you are assigning `@ foods` to the entire JSON response, which has A LOT more data than what we need to send to the view. So the facade would be responsible for taking the response from the service, iterating through the foods collection that's returned in the JSON, and creating plain old ruby objects (POROs) out of the collection, to then be stored into this instance variable.

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,13 @@
...
+<section>
+  Number of search results: <%= @foods[:totalHits] %><br>
+  <% @foods[:foods].each do |food| %>
+    <section id="food-<%= food[:fdcId] %>">
+      GTIN/UPC: <%= food[:gtinUpc] %><br>
+      Description: <%= food[:description] %><br>
+      Brand Owner: <%= food[:brandOwner] %><br>
+      Ingredients: <%= food[:ingredients] %><br><br>
+    </section>
+  <% end %>
```
megstang, line 12: Just to reiterate what I mentioned above, `@ foods` is your entire JSON response from your api call. This is unnecessary to send all of this information over, and it can also be dangerous, if there is data in that response that you don't want your user's to have access to.

megstang (commented):
@A-McGuire Hey Aidan, I'm going to close this out and follow up in a DM with your scores shortly.

PR #12, opened 2021-05-24
Jahara Clark

jaharaclark:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,6 @@
+class FoodsController < ApplicationController
+   def index
+    food_input = 'sweet potatoes'
+    @food_response = FoodFacade.foods(food_input)
```
megstang, line 4: Careful with hardcoding your search. This search string should come from the search form on your welcome index page upon submission.

app/services/food_service.rb
```diff
@@ -0,0 +1,18 @@
...
+    response  =  conn.get("fdc/v1/foods/search?query=#{food_input}")
+    parse_response(response)
+  end
+
+  private
+  def self.conn
+    conn = Faraday.new("https://api.nal.usda.gov") do |faraday|
+         faraday.params['api_key'] = ENV['food_token']
+       end
+  end
```
megstang, line 12: You do not need to save this into a variable on line 9. This whole block is essentially the last line of your method, so this connection will get returned without declaring the variable.

app/services/food_service.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def self.parse_response(response)
+    JSON.parse(response.body, symbolize_names: true)[:foods]
+  end
```
megstang, line 16: There is still more we'd like to get out of this response-- we also want to know the total number of hits that we got from the search. By selecting the value that came from the `:foods` key, you've lost the totalhits key/value.

app/facades/food_facade.rb
```diff
@@ -0,0 +1,5 @@
+class FoodFacade
+  def self.foods(food_input)
+    food_result = FoodService.return_food_query(food_input)
```
megstang, line 3: This method is returning the entire JSON response to the controller, which will then send it to the view. There's A LOT in this JSON response that we don't need to send to the view-- there's a lot of food (of which we'll only select 10 of), and within each food there's a lot of data that we (at this moment) don't care to share with our client. This is where the PORO comes in.. let's create Food objects out of the first 10 foods, and THEN send that collection to the view.

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,10 @@
+<h1>Food Search Results</h1>
+<% @food_response.first(10).each do |food|%>
+<section class='food_item'>
+  <h2 class='gtinUpc'>Food GTIN/UPC: <%=food[:gtinUpc]%></h2>
+  <h2 class='description'>Food Description: <%=food[:description]%></h2>
+  <h2 class='brandOwner'>Food Brand Owner: <%=food[:brandOwner]%></h2>
+  <h2 class='ingredients'>Food Ingredients: <%=food[:ingredients]%></h2>
+</section>
```
megstang, line 8: So here, `@food_response` is the nearly the entire JSON response. When you refactor your code to use a PORO, then you'd be able to iterate through food ruby objects.

megstang (commented):
@jaharaclark I left a few comments for you to look over. Let me know if you have any questions, please. I'm going to follow up via slack to send over your scores as well as some more feedback.

PR #14, opened 2021-05-24
2102 Mid Mod - Mike Foy

foymikek:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,12 @@
+class FoodsController < ApplicationController
+  def index
+    food_input          = params[:q]
+    if food_input       == ""
+     flash[:alert]      = "Please input a food item to search!"
+     redirect_to root_path
```
timomitchel, line 6: Love the sad path implementation here. Another protection I usually use is `.nil?`. And, for the comparison you're doing here, it's best to use `.blank?` instead of `== ""`

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,12 @@
...
+    else
+      @foods_info       = FoodsFacade.search_by_food(food_input)
+      @total_food_count = FoodsFacade.searched_food_count(food_input)
```
timomitchel, line 9: It's totally fine to pass two instance variables to the view in this case, but it would be best to only make one request to the API, because we can get all the info we need with one request.

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,15 @@
...
+      searched_food_info = FoodsService.search_by_food_db(food_input)
+
+      searched_food_info[:foods].map do |food_info|
+        Food.new(food_info)
+      end
+    end
+
+    def searched_food_count(food_input)
+      FoodsService.search_by_food_db(food_input)[:totalHits]
+    end
```
timomitchel, line 13: Instead of having two methods here we could do something like this:
```ruby
searched_food_info = FoodsService.search_by_food_db(food_input)
food_data = searched_food_info[:foods].map do |food_info|
        Food.new(food_info)
      end
return [searched_food_info[:totalHits], food_data]
```

This way we return a 2d array with everything we need back to the controller. Then, we can unpack this array like so: 
```ruby
@total_food_count, @foods_info = FoodsFacade.search_by_food(food_input)
```
That's some fancy syntax there!

app/poros/food.rb
```diff
@@ -0,0 +1,13 @@
+class Food
```
timomitchel, line 1: Nice simple poro

app/services/foods_service.rb
```diff
@@ -0,0 +1,20 @@
+class FoodsService
+  class << self
+    def search_by_food_db(food_input)
+      response =  conn.get("fdc/v1/foods/search?query=#{food_input}")
```
timomitchel, line 4: This class is excellent, the only thing I would change is I would set the `query` param in a block the same way that you set the API key down in the `conn` method.

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,17 @@
+Searched item count: <%= @total_food_count %>
+<% @foods_info[0..9].each do |food_info| %>
```
timomitchel, line 2: This `@foods_inof[0..9]` is breaking the SRP of the view and it shows our code is inefficient. Essentially, we only want to ever make 10 food objects, so this should happen in the facade like so:
```ruby
searched_food_info[:foods].first(10).map do |food_info|
        Food.new(food_info)
      end
```
This way we aren't creating unused objects.

spec/facades/foods_facade_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'rails_helper'
+
+describe FoodsFacade do
+  it 'exists' do
```
timomitchel, line 4: Great job testing here.

spec/services/foods_service_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'rails_helper'
+
+describe FoodsService do
+  it 'exists' do
```
timomitchel, line 4: Another excellent test file.

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+      fill_in :q, with: "sweet potatoes"
+      click_button "Search"
+
+      expect(current_path).to eq(foods_path)
+
+
+      expect(page).to have_content("Searched item count: 44128")
+
+      expect(page).to have_css('.food_information', count: 10)
```
timomitchel, line 16: Excellent tests that follow the spec exactly. However the actual count of the `Searched Item Count` could break anytime the API changes. It would have been best to stub this request with data you can depend on remaining consistent.

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+      within(first('.food_information')) do
+        expect(page).to have_css('.gtin_upc_code')
+        expect(page).to have_css('.description')
+        expect(page).to have_css('.brand_owner')
+        expect(page).to have_css('.ingredients')
+      end
+    end
+
+    it "I am prompted to enter a valid search if no search is submitted" do
```
timomitchel, line 27: Nice sad path testing.

timomitchel (commented):
Overall, a stellar job on this challenge. LMK if you have questions on the feedback.

PR #15, opened 2021-05-24
Andrew Shafer's Work

Aphilosopher30:


Aphilosopher30:
where I am: I got some of the basic functionality, but i didn't get to make any features, poros, or sad paths due to time constraints, although i think i have a good idea on how to implement those if i had time 

 I feel i understand the basic principles. but some strange ness towards the end makes me think that the API might now be working how it though it was, so that might be something of a problem

also, i don't know the stuff well enough to make a good test from the start, before writing the code. so i typically make a bad, or even broken test, then write the code, and then figure out what i need to make a good test.  but if the point of a test is to make something that checks that it is doing what it is supposed to do, then i can't currently do that untill i have a clearer idea of what it is that my code is supposed to do, and so i must write the code before i can write a good test. that's where i am in my learning.

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,10 @@
+class FoodsController < ApplicationController
+    def index
+      @ingredient = params[:q]
+      @food_count = FoodsService.search_ingredients(@ingredient)[:totalHits]
+      # binding.pry
+      @foods = FoodsService.food_list(@ingredient) 
+      # binding.pry
+
+    end
```
megstang, line 9: Ideally, if we can have one object getting sent to the view, we should try to do that. The one object will contain all data necessary for displaying in the view. You mentioned in a comment that you would Ideally call a Facade here and then let your method in the Facade call your Service, and I would agree with that refactor. We shouldn't have to do any manipulation/formatting of our data in the controller.

app/services/foods_service.rb
```diff
@@ -0,0 +1,39 @@
+class FoodsService
+
+  def self.search_ingredients(ingredient)
+    paramiters = {query: ingredient}
+    response = connection.get('fdc/v1/foods/search') do |faraday|
+      paramiters.each do |k, v|
+        faraday.params[k] = v
+      end
+    end
```
megstang, line 9: This api also can take an optional parameter of `per_page` which you could leverage to get only 10 foods back since the default is to just send the first page.

app/services/foods_service.rb
```diff
@@ -0,0 +1,39 @@
...
+    # binding.pry
+    future_poro = top_ten.map do |food|#in future make this poro
+      {brand: food[:brandOwner],
+      description: food[:lowercaseDescription],
+      gtinUpc: food[:gtinUpc],
+      ingredients: food[:ingredients]}
+    end
+    future_poro
+
+  end
```
megstang, line 27: Thanks for putting these notes in here -- helps me to see how you would ideally refactor it had you had more time.

megstang: You could also grab the totalHits from `food_data`, and put that into the object you're returning to your controller.. that would help eliminate the amount of instance variables you're currently sending to the view.

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+
+
+    fill_in 'q', with: "sweet potato"
+    click_button "Search"
+    expect(current_path).to eq("/foods")
+
+    search = "sweet potato"
+
+    expect(page).to have_content("total foods with #{search}: #{44128}")
```
megstang, line 41: You don't need to interpolate the amount here

megstang (commented):
Hey @Aphilosopher30 I left a few comments for you to look over. I'll follow up with you in a DM to provide some more feedback as well as your scores.

PR #16, opened 2021-05-24
Ryan Stemmle

r-stemmle:
Ran into lots of small errors that ate my time up..

Api key was showing nil forever and then it just started working? ha

Didn't put my feature test in a features folder and rpsec did not like that for about 10 minutes

Not even sure my query is finding exactly what the user story wants

Ryan


timomitchel:
Would have loved to see more commits throughout this assessment.

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,11 @@
+class FoodsController < ApplicationController
+  API = ENV['fdc_api_key']
```
timomitchel, line 2: You shouldn't have to set this constant if you configure figaro appropriately.

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,11 @@
...
+
+  def index
+    url = 'https://api.nal.usda.gov/fdc/v1/foods/search'
+    response = Faraday.get(url, { query: params["q"], api_key: API }, {'Content-Type' => 'application/json'})
+    resp = JSON.parse(response.body, symbolize_names: true)
+    @count = resp[:totalHits]
+    @foods = resp[:foods].first(10)
```
timomitchel, line 9: This all looks pretty good for a starting point. One thing to consider is setting your params and headers in a block to make this a bit more readable/maintainable.

app/models/food.rb
```diff
@@ -0,0 +1,2 @@
+class Food < ApplicationRecord
```
timomitchel, line 1: This file can be deleted.

app/poros/food.rb
```diff
@@ -0,0 +1,10 @@
+class Food
+  attr_reader :gtin_ucp_code, :description, :brand, :ingredients
+  def initialize(data)
+    @gtin_ucp_code = data[:gtinUpc]
+    @description = data[:lowercaseDescription]
+    @brand = data[:brandOwner]
+    @ingredients = data[:ingredients]
+  end
```
timomitchel, line 8: This is great, exactly what I would expect to see in this object.

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,26 @@
...
+    <% @foods.each do |food| %>
+      <tr>
+        <td><%= food[:gtinUpc] %></td>
+        <td><%= food[:lowercaseDescription] %></td>
+        <td><%= food[:brandOwner] %></td>
+        <td><%= food[:ingredients] %></td>
+      </tr>
+    <% end %>
+  </tbody>
+</table>
```
timomitchel, line 26: I like this setup, but I don't want you to spend extra time on a timed assessment formatting the view unless the spec asks for that.

config/routes.rb
```diff
@@ -1,3 +1,5 @@
 Rails.application.routes.draw do
   root "welcome#index"
+
+  resources :foods
```
timomitchel, line 4: Would have loved to see an `only: [index]` here so we aren't creating extra routes.

db/migrate/20210524151343_create_foods.rb
```diff
@@ -0,0 +1,11 @@
+class CreateFoods < ActiveRecord::Migration[5.2]
```
timomitchel, line 1: No need for this migration/table. Our poros can handle this, and we don't want to store info from the API unless there is no way around that.

spec/features/welcome_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+      click_on 'Search'
+
+      expect(current_path).to eq(foods_path)
+      expect(page).to have_content('Foods')
+      expect(page).to have_content('RESULTS')
+      expect(page).to have_content('Gtin upc code')
+      expect(page).to have_content('Description')
+      expect(page).to have_content('Brand')
+      expect(page).to have_content('Ingredient')
+    end
```
timomitchel, line 18: Would have also been a good idea to count the `<tr>` elements to ensure there were only 10 of them. As well as the actual number of total results and some of the specific food data we expect to be on the page.

timomitchel (commented):
Functionality is here, but looks like you weren't able to refactor or completely test your functionality. This is all about practice. Make sure to continue to get some reps on API consumption, and I would recommend finishing the refactor and adding to the test suite at some point this week for practice.

PR #17, opened 2021-05-24
Gaelyn Cooper

gaelyn:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,6 @@
+class FoodsController < ApplicationController
+  def index
+    @results = FoodFacade.total_results(params[:q])
+    @foods = FoodFacade.find_foods(params[:q])
```
megstang, line 4: Since both `total_results` and `find_foods` are both calling `search_by_ingredient` (they're hitting the same endpoint, with the same exact query params), I'd like to challenge you to find a way that you could only make that call once. Currently, you're calling to the same endpoint twice, and getting back the same data both times, but then grabbing a more specific part of the response in your facade to return back to this controller. 

See if you can find a way to just call `find_foods`, and have that method return both the total hits and the food collection.

megstang: You could take a look at this syntax:

`results, foods= FoodFacade.your_method(params[:q])`

Where your_method would return a 2d array with the number of results as the first element and the food collection as the second element.

megstang (commented):
@gaelyn Good job, I left a few comments for you to look over. Let me know if you have any questions. I'll follow up shortly via slack to send over your scores.

PR #18, opened 2021-05-24
Klaudia Stewart - Mid mod

klaudiastewart:
I think this went well. For whatever reason, my entire rspec suite does not run all at once, so I need to run individual files. The vcr cassettes are a bit strange too. Not sure why I have to delete the cassette to run a fresh spec. I am guessing something is up with my vcr set-up. Some feedback on this would be nice. Other than that, I believe that I encapsulated and abstracted things in a good manner, and the feature functionality is fully there. Again, tests may have fallen just a bit short, but maybe you think otherwise.


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,8 @@
+class FoodsController < ApplicationController
+  require "./app/facades/food_facade.rb"
```
timomitchel, line 2: No need for this `require`-  everything placed in `/app/` is available.

app/poros/food_item.rb
```diff
@@ -0,0 +1,13 @@
+class FoodItem
+  attr_reader :code,
```
timomitchel, line 2: Simple and effective PORO here.

app/facades/food_facade.rb
```diff
@@ -0,0 +1,10 @@
+class FoodFacade
+  require "./app/services/food_service.rb"
+
+  def self.get_food_items(food)
+    items = FoodService.call_for_items(food)
+    items[:foods][0..9].map do |data|
```
timomitchel, line 6: Glad to see you only ever creating 10 FoodItem objects. This is efficient and how it should be done. I would break up the `map` and the `items` call to the service into two methods, so that I could thoroughly test both.

app/services/food_service.rb
```diff
@@ -0,0 +1,21 @@
...
+    def call_for_items(food)
+      response = conn.get("/fdc/v1/foods/search?query=#{food}")
+      parse_data(response)
+    end
+
+    private
+
+    def conn
+      Faraday.new(
+        url: "https://api.nal.usda.gov",
```
timomitchel, line 12: Love seeing you set these on separate lines to increase readability.

app/services/food_service.rb
```diff
@@ -0,0 +1,21 @@
+class FoodService
+  class << self
+    def call_for_items(food)
+      response = conn.get("/fdc/v1/foods/search?query=#{food}")
```
timomitchel, line 4: Similar to your `conn` method, I would set `query` in a block or on a separate line to add readability/maintainability.

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,12 @@
+<h2>Total number of items with sweet potatoes: <%= %></h2>
```
timomitchel, line 1: This was a bit tricky to figure out. But, there was a key in the JSON response called `totalHits` that could have been passed from the facade >controller > view.

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,12 @@
...
+<h2>Ten foods that contain sweet potatoes:</h2>
+<ul>
+  <% @foods.each do |food| %>
+  <li>
+      Code: <%= food.code %> |
+      Description: <%= food.description %> |
+      Brand Owner: <%= food.brand_owner %> |
+      Ingredients: <%= food.ingredients %>
+    <% end %>
+  </li>
```
timomitchel, line 11: I would probably add some classes to these elements so that you can utilize `within` blocks to target specific elements/areas and count that the number of elements on the page is indeed correct.

spec/facades/food_facade_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'rails_helper'
+
+RSpec.describe 'FoodFacade' do
+  it "returns an array of food objects" do
+    stub_food_request
```
timomitchel, line 5: Love that you used a helper method and stubbed the requrest. However, I would want this testing to be a bit more thorough, which could be accomplished with sad paths and breaking up your facade into 2 methods.

spec/features/search_for_sweet_potatoes.rb
```diff
@@ -0,0 +1,12 @@
...
+
+RSpec.describe 'searching for food', type: :feature do
+  it 'can search for sweet potatoes' do
+    visit root_path
+
+    fill_in :q, with: 'sweet potatoes'
+    click_button 'Search'
+
+    expect(current_path).to eq(foods_path)
+  end
```
timomitchel, line 11: This test is redundant and can be removed.

spec/spec_helper.rb
```diff
@@ -94,3 +95,15 @@
   Kernel.srand config.seed
 =end
 end
+
+def stub_food_request
```
timomitchel, line 99: Love this helper method. However, it doesn't look like you ever utilize VCR, because you did not add the configuration block, and you never use the vcr syntax in any of your tests to use the cassette.

spec/features/items_foods_show_page_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'rails_helper'
+
+RSpec.describe 'show page for the searched food', type: :feature do
```
timomitchel, line 3: Would also have loved to see a service and PORO test.

timomitchel (commented):
Overall a good job on this mid mod. You're showing a ton of progress. LMK if you have questions on the feedback.

PR #19, opened 2021-05-24
Brisa Submission

brisag:
I don't think i know how to use facades that well.  I thought you call the method from the service file to iterate through a PORO.new(params)  and it works.  Not sure what went wrong!

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -4,12 +4,3 @@ class SearchController < ApplicationController
     @food_count = FoodFacade.count_foods_containing(params[:q])
   end
 end
-    # food_input = 'sweet potatoes'
-    # conn      = Faraday.new("https://api.nal.usda.gov") do |faraday|
-    #   faraday.params['api_key']
-    # end
-    # response  =  conn.get("fdc/v1/foods/search?query=#{food_input}")
-    # @food_info = JSON.parse(response.body, symbolize_names: true)[:foods]
-    # @food_info.map do |info|
-    #   # require 'pry'; binding.pry
-    #   info
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -1,0 +1,23 @@
+class SearchController < ApplicationController
+  def index
+    @food_info = FoodFacade.foods_containing(params[:q])
+    @food_count = FoodFacade.count_of_foods_containing(params[:q])
+  end
+end
+    # food_input = 'sweet potatoes'
+    # conn      = Faraday.new("https://api.nal.usda.gov") do |faraday|
+    #   faraday.params['api_key'] = '5W96aX0rTBARrDA2jxr5SsBZNJghEjbddhV6xCzI'
+    # end
+    # response  =  conn.get("fdc/v1/foods/search?query=#{food_input}")
+    # @food_info = JSON.parse(response.body, symbolize_names: true)[:foods]
+    # @food_info.map do |info|
+    #   # require 'pry'; binding.pry
+    #   info
+
+    # @foods = FoodFacade.foods_containing(params[:q])
+    # @foods_count = FoodFacade.count_of_foods_containing(params[:q])
+  #   keyword = params[:query]
+  #   conn = Faraday.new('https://api.nal.usda.gov/fdc')
+  #   response = conn.get("/v1/foods/search?query=#{keyword}")
+  #   @food_infos = JSON.parse(response.body, symbolize_names: true)[:foods]
+  # end
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -1,23 +1,15 @@
 class SearchController < ApplicationController
   def index
     @food_info = FoodFacade.foods_containing(params[:q])
-    @food_count = FoodFacade.count_of_foods_containing(params[:q])
+    @food_count = FoodFacade.count_foods_containing(params[:q])
   end
 end
     # food_input = 'sweet potatoes'
     # conn      = Faraday.new("https://api.nal.usda.gov") do |faraday|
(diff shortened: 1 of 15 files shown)


brisag:
I went back to fix my refactored faraday connection

app/controllers/search_controller.rb
```diff
@@ -0,0 +1,6 @@
+class SearchController < ApplicationController
+  def index
+    @food_info = FoodFacade.foods_containing(params[:q])
+    @food_count = FoodFacade.count_foods_containing(params[:q])
```
megstang, line 4: Both of these methods, `foods_containing` and `count_foods_containing` call the same endpoint, and just pull different data. I'd like to challenge you to find a way to just call one method in your facade, and have your facade take care of combining the food objects and the totalHits into one object to get returned to your controller. All of the data that you need for this view can come from the one query call you're making.

megstang: You could look at using syntax like this: 

`food_count , food_info = FoodFacade.your_method(params[:q])`
Where `your_method` will return a 2d array with the first element being the food count, and the second element being the food object collection.

app/facades/food_facade.rb
```diff
@@ -0,0 +1,14 @@
+class FoodFacade
+  class << self
+    def foods_containing(keyword)
+      # binding.pry
+      FoodService.foods_containing(keyword)[:foods].each do |food_info|
+        Food.new(food_info)
+      end
```
megstang, line 7: I think you may have gotten "each'd" 😬 , which is why you may have thought it wasn't working. If you swapped `each` out for a better case enumerable, I think this code should work.

brisag: I got it to work! i made a mistake in my service file!

brisag: My original response was missing the fdc.  I was moving very fast.
response = conn.get("fdc/v1/foods/search?query=#{keyword}")

megstang: Going off of what I mentioned in your controller regarding avoiding calling two different methods that make the same api call: 

You could assign the food collection into a variable, let's say `food_collection` and the total foods count to a variable, let's say `food_count`, and then have your method return `[food_count, food_collection]`.. this return value would then be saved into the respective variables in your controller, to the left of the = sign.

app/services/food_service.rb
```diff
@@ -0,0 +1,20 @@
+class FoodService
+  class << self
+    def foods_containing(keyword)
+      response = conn.get("fdc/v1/foods/search?query=#{keyword}")
+      parse_data(response)
+    end
```
megstang, line 6: you could utilize an optional parameter from the api, `per_page` to get exactly 10 results back

spec/features/index_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      click_on "Search"
+      expect(current_path).to eq(foods_path)
+    end
+
+    it ' should see a total of the number of items returned by the search (more than 30,000 results)' do
+      visit root_path
+
+      fill_in :q, with: "sweet potatoes"
+      click_on "Search"
+      expect(current_path).to eq(foods_path)
```
megstang, line 17: you don't need to test this since it's in your last test

spec/features/index_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+        within(first(".foods")) do
+          expect(page).to have_content("GTIN/UPC code: ")
+          expect(page).to have_css(".code")
+          expect(page).to have_content("Description: ")
+          expect(page).to have_css(".description")
+          expect(page).to have_content("Brand Owner: ")
+          expect(page).to have_css(".brand_owner")
+          expect(page).to have_content("Ingredients: ")
+          expect(page).to have_css(".ingredients")
+        end
```
megstang, line 39: you can nest within's too -- within .foods, you could check within .code for whatever you'd expect to be there.

app/services/food_service.rb
brisag: Yea! I will look into pagination! I did considered putting in the facade.  i know have it locally as: 

  def foods_containing(keyword)
      foods = FoodService.foods_containing(keyword)
      # binding.pry
      foods[:foods][0..9].each do |food_info|
        Food.new(food_info)
      end
    end

spec/poros/food_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    food_info = Food.new(attributes)
+
+    expect(food_info).to be_an_instance_of(Food)
+    expect(food_info.upc).to eq("492111402857")
+    expect(food_info.description).to eq("sweet potatoes")
+    expect(food_info.brand_owner).to eq("ARCHER FARMS")
+    expect(food_info.ingredients).to eq("SWEET POTATOES.")
+  end
+end
```
megstang, line 20: I like that you remembered to test your PORO. From what I see, the argument you pass to the Food class's initialize method is a lot bigger than the example you have on lines 5-10. It may be good to get some fixture data from one food element in the JSON response, and use that here in the test.

megstang (commented):
@brisag Really good job. I left a few comments for you to look over -- I think I was able to find why it wasn't working for you, and it should just be a small fix. I'll follow up shortly with you regarding scores.

app/facades/food_facade.rb
megstang: I definitely didn't even catch that, but glad you did! -- I think you'll still want to change it to a `map` though. In your view, you're accessing the JSON object since the JSON response is what's getting returned from your `foods_containing` method in your FoodFacade. Currently, in your view, since you're not using `map`, you're not working with food objects, but the data is still there because you have access to the entire response. Let me know if I can be more clear with what I'm saying, and please let me know if you have any follow up questions.

PR #21, opened 2021-05-24
Diana Buffone

Diana20920:
I can tell I have improved a lot from the last mid mod I did during my first round of mod 3. 
Not sure how or if feedback will be provided, but I would really appreciate any that ya'll can provide so I can continue to solidify my API consumption skills. 
Thank you! 😃

### DIFF
--- a/app/controllers/foods_controller.rb
+++ b/app/controllers/foods_controller.rb
@@ -7,6 +7,9 @@ class FoodsController < ApplicationController
 
     @total_results = parsed[:totalHits]
 
-    @foods = parsed[:foods].first(10)
+    # @foods = parsed[:foods].first(10)
+    @foods = parsed[:foods].map do |food|
+      Food.new(food)
+    end.first(10)
   end
 end
--- a/app/controllers/foods_controller.rb
+++ b/app/controllers/foods_controller.rb
@@ -8,8 +8,5 @@ class FoodsController < ApplicationController
     @total_results = parsed[:totalHits]
 
     @foods = parsed[:foods].first(10)
-
-
-    binding.pry
   end
 end
--- a/app/controllers/foods_controller.rb
+++ b/app/controllers/foods_controller.rb
@@ -1,15 +1,7 @@
 class FoodsController < ApplicationController
   def index
-    query = params[:q]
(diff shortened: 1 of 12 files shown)


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,9 @@
+class FoodsController < ApplicationController
+  def index
+    @searched_ingredient = params[:q]
+    
+    @total_results = FoodFacade.total_results_ingredient(@searched_ingredient)
+
+    @foods = FoodFacade.foods_with_ingredient(@searched_ingredient)
+  end
```
megstang, line 8: You're code's looking good.. so... a little challenge for ya 😏 : 

Both of these methods, `total_results_ingredient` and `foods_with_ingredient` call the same endpoint, and just pull different data. I'd like to challenge you to find a way to just call one method in your facade, and have your facade take care of combining the food objects and the totalHits into one object to get returned to your controller. All of the data that you need for this view can come from the one query call you're making.

megstang: You could take a look at this syntax:

`@ total_results, @ foods = FoodFacade.your_method(params[:q])`

Where your_method returns a 2d array, with the number of results as the first element and the food collection as the second element.

app/facades/food_facade.rb
```diff
@@ -0,0 +1,13 @@
+class FoodFacade
+  def self.foods_with_ingredient(ingredient)
+    json = FoodService.ingredient_query(ingredient)
+    json[:foods].map do |food|
+      Food.new(food)
+    end.first(10)
```
megstang, line 6: You could look into pagination with your request so you don't need to limit it here -- a plus side of using pagination to limit the results you get back is that you won't need to iterate through ALL of the foods that are returned from the call.

megstang: And going off of the comment/challenge I put on your controller above, if you could get this method to return both the total hits and the collection of food to the controller, that could also limit how many times you are calling the same endpoint.

megstang (commented):
@Diana20920 Good job, I left a few comments for you to look over. Please let me know if oyu have any questions. I'll follow up shortly with your scores over slack.

PR #22, opened 2021-05-24
Kudzelich PR

gitkseniya:


app/facades/food_facade.rb
```diff
@@ -0,0 +1,15 @@
...
+    FoodService.get_details(food)
+  end
+
+  def self.food_info(info)
+    food = service(info)
+
+    food.map do |food_data|
+      Food.new(food_data)
+    end
+  end
```
timomitchel, line 13: This looks mostly good, but we don't have a way to extract the `totalResults` here, which is the responsibility of the facade.

app/facades/food_facade.rb
```diff
@@ -0,0 +1,15 @@
+class FoodFacade
+
+  def self.service(food)
+    FoodService.get_details(food)
+  end
+
+  def self.food_info(info)
+    food = service(info)
+
+    food.map do |food_data|
```
timomitchel, line 10: Would have liked to see you do a `food.first(10).map...` so that we only ever create 10 objects instead of thousands. This is a big performance hit.

app/poros/food.rb
```diff
@@ -0,0 +1,14 @@
+class Food
+  attr_reader  :upc,
```
timomitchel, line 2: Simple and effective PORO.

app/services/food_service.rb
```diff
@@ -0,0 +1,32 @@
+class FoodService < BaseService
```
timomitchel, line 1: Really love you utilizing inheritance here.

app/services/food_service.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    response =  conn("https://api.nal.usda.gov").get("/fdc/v1/foods/search") do |f|
+      f.params["query"] = food
+      f.params["api_key"] = ENV['api_key']
+      f.params["pageSize"] = 200
+      # f.params["pageNumber"] = page_number
+
+      if !f.params["pageNumber"] == 5
+        page_num += 1
+      end
```
timomitchel, line 15: No need for any pagination within the context of the stories. Love seeing you set your prams in the block. The API key should be set within the `conn` method in `BaseService`

app/services/food_service.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    end
+
+    json = format_json(response)
+    json[:foods]
```
timomitchel, line 20: This is data massaging. This belongs in the facade.

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,8 @@
+Total number of items: <%= @foods.count %><br><br>
```
timomitchel, line 1: I see why you created all these food objects now. But, there is a `totalHits` key returned from the API  call that we can extract this information from in a more efficient manner.

config/routes.rb
```diff
@@ -1,3 +1,4 @@
 Rails.application.routes.draw do
   root "welcome#index"
+  resources :foods, only: [:index]
```
timomitchel, line 3: Nice use of `only` to only create the necessary route.

spec/facades/food_facade_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'rails_helper'
+
+RSpec.describe 'FoodFacade' do
+  it "returns an array of food objects", :vcr do
```
timomitchel, line 4: Glad to see you using vcr here.

spec/facades/food_facade_spec.rb
```diff
@@ -0,0 +1,11 @@
...
+
+    facade = FoodFacade.service("sweet potatoes")
+
+    expect(facade).to be_an Array
+    expect(facade.first).to be_a Hash
```
timomitchel, line 9: These expectations could be more robust. Would have also liked to see the `food_info` method unit tested.

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+
+      expect(current_path).to eq(foods_path)
+      expect(page).to have_content('Total number of items:')
+
+      expect(page).to have_content("832298010009")
+      expect(page).to have_content("C.H. Robinson Company
+")
+      expect(page).to have_content("NOT A BRANDED ITEM")
+      expect(page).to have_content("Spring Acres Sales Company Inc.")
+    end
```
timomitchel, line 20: Again, a pretty solid test, but it could be more robust. Counting the number of css_selectors should be 10. Making sure the actual number of results reflected is what you expect from your stub, etc.

spec/poros/food_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+
+
+
+
+  expect(food.upc).to eq "492111402857"
+  expect(food.description).to eq "Potato, mashed, from restaurant"
+  expect(food.brand_owner).to eq "ARCHER FARMS"
+  expect(food.ingredients).to eq "SWEET POTATOES."
+  end
```
timomitchel, line 33: Simple effective test.

spec/services/food_service_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'rails_helper'
+
+RSpec.describe 'Food API' do
+  it 'returns food data' do
+    food_data = FoodService.get_details("sweet potatoes")
+
+    expect(food_data).to be_an(Array)
+    expect(food_data.first).to be_a(Hash)
```
timomitchel, line 8: This is the proper place to test the structure of your API response and the data types. These tests can then be removed from the facade test, and instead we should test a pure input/output that returns the exact value we expect.

spec/services/food_service_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    sweet_potato = food_data[0]
+
+    expect(sweet_potato[:gtinUpc]).to eq("492111402857")
+    expect(sweet_potato[:foodCategory]).not_to eq("070038322238")
+    expect(sweet_potato[:brandOwner]).not_to eq(sweet_potato[:brandName])
```
timomitchel, line 14: No need for these tests at the service level. This is better tested at the PORO level, which you did, so this is redundant.

timomitchel (commented):
Overall a great job on this mid mod. LMK if you have questions on the feedback.

PR #23, opened 2021-05-24
Joe Mecha

joemecha:
User story completed

Spent time tracking down unneeded '@' in view causing nil error

As a result, all API call code is still in the controller. No time remaining for refactor into service/facade

### DIFF
--- a/app/controllers/foods_controller.rb
+++ b/app/controllers/foods_controller.rb
@@ -1,6 +1,6 @@
 class FoodsController < ApplicationController 
   def index
-    @query = params[:q]  
+    @query = params[:q]
     key = "X5MqWQCLRSlxUlXnpdm2q51W13hIhOhj5krsN0ep"
     conn = Faraday.new(url: "https://api.nal.usda.gov") do |faraday|
       faraday.headers['api_key'] = "X5MqWQCLRSlxUlXnpdm2q51W13hIhOhj5krsN0ep" #ENV['usda_api_key']
@@ -8,10 +8,9 @@ class FoodsController < ApplicationController
     response = conn.get("/fdc/v1/foods/search?query=#{@query}&numberOfResultsPerPage=10&api_key=#{key}")
 
     json = JSON.parse(response.body, symbolize_names: true)
+    @num_results = json[:totalHits]
     @foods = json[:foods].map do |food_data|
       Food.new(food_data)
-    end 
-
-    # need to get num of results somehow
+    end
   end
 end
--- a/app/controllers/foods_controller.rb
+++ b/app/controllers/foods_controller.rb
@@ -1,7 +1,13 @@
 class FoodsController < ApplicationController 
   def index
-    require 'pry'; binding.pry
-    query = params[:q]  
-  end
+    @query = params[:q]  
+    conn = Faraday.new(url: "https://api.nal.usda.gov") do |faraday|
+      faraday.headers['api_key'] = "X5MqWQCLRSlxUlXnpdm2q51W13hIhOhj5krsN0ep" #ENV['usda_api_key']
+    end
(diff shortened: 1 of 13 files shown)


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,19 @@
+class FoodsController < ApplicationController 
+  def index
+    @query = params[:q]
+    key = "X5MqWQCLRSlxUlXnpdm2q51W13hIhOhj5krsN0ep"
+    conn = Faraday.new(url: "https://api.nal.usda.gov") do |faraday|
+      faraday.headers['api_key'] = "X5MqWQCLRSlxUlXnpdm2q51W13hIhOhj5krsN0ep" #ENV['usda_api_key']
+    end
+    
+    response = conn.get("/fdc/v1/foods/search?query=#{@query}&numberOfResultsPerPage=10&api_key=#{key}")
```
timomitchel, line 9: Would love to see the params set in a block so that this is more readable. Also, not good that you pushed up your API key to github. That's a security vulnerability that can have your permissions from an API revoked. Looks like you set it both as a header and as a param. Generally, you only set that in one of those places depending on what the documentation asks for. Again, this could cause a security vulnerability.

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    json = JSON.parse(response.body, symbolize_names: true)
+    
+    @num_results = json[:totalHits]
+    
+    @foods = json[:foods].map do |food_data|
```
timomitchel, line 15: Would love to see you scope this to the `.first(10).map...` so that we only ever create 10 Food objects instead of thousands. This currently causes a performance issue.

app/poros/food.rb
```diff
@@ -0,0 +1,13 @@
+class Food 
+  attr_reader :upc_code,
```
timomitchel, line 2: Simple and effective Food object.

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,15 @@
+<h1>Foods Search Result</h1>
+
+<b>Total items returned: <%="#{@num_results}"%></b>
+
+<h3>Foods containing <%=@query%></h3>
+
+<% @foods[0..9].each do |f| %>
```
timomitchel, line 7: Not where we want to scope this to 10 as I described why above.

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+      expect(current_path).to eq(foods_path)
+      
+      expect(page.status_code).to eq 200
+      expect(page).to have_content("Total items returned: 44128")
+      expect(page).to have_content("GTIN/UPC code:")
+      expect(page).to have_content("Description:")
+      expect(page).to have_content("Brand Owner:")
+      expect(page).to have_content("Ingredients:")
+    end
```
timomitchel, line 19: Solid test, but could be more robust and improved with `within` blocks, counting the number of `Food` elements on the page which should be 10, and thinking about a sad path. Also, we should be stubbing the API call.

spec/poros/food_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+                    "value": 0.0
+                }
+            ]
+        }
+      
+    food =  Food.new(food_data)
+    expect(food.upc_code).to eq "492111402857"
+    expect(food.description).to eq "SWEET POTATOES"
+    expect(food.brand_owner).to eq "ARCHER FARMS"
+    expect(food.ingredients).to eq "SWEET POTATOES."
```
timomitchel, line 151: Solid testing, but better to read that json data from a file to help this test be more readable/maintainable.

timomitchel (commented):
Overall a good job on this mid mod. Just need to speed up a bit, but on the right track. LMK if you have questions on the feedback.

PR #24, opened 2021-05-24
Keegan O'Shea

ghost:
I ran out of time refactoring to a facade architecture but overall felt ok about this. I was leaning on the feature test during the refactor, but would have liked to have added a test for the service, facade and features.

Thanks for your time!

### DIFF
--- a/app/facades/foods_facade.rb
+++ b/app/facades/foods_facade.rb
@@ -1,0 +1,19 @@
+class FoodsFacade
+  def self.search_food(query)
+    foods = []
+    data = FoodsService.search_food(query)
+    data[:foods].first(10).each do |food|
+      foods << Food.new(food)
+    end
+  end
+
+  def self.search_food_query(query)
+    data = FoodsService.search_food(query)
+    data[:foodSearchCriteria][:query]
+
+  def self.search_food_total(query)
+    data = FoodsService.search_food(query)
+    data[:totalHits]
+    end
+  end
+end
--- a/app/services/foods_service.rb
+++ b/app/services/foods_service.rb
@@ -1,0 +1,22 @@
+class FoodsService
+
+  def self.conn
+      Faraday.new(
+      url: "https://api.nal.usda.gov/fdc/",
+      params: {},
+      headers: {"X-Api-Key" => "#{ENV["FOOD_DATA_API_KEY"]}"
+    }
+    )
+  end
+
+  def self.search_food(query)
+    response = conn.get("v1/foods/search?query=#{query}")
+    parse_json(response)
+  end
+
+  private
+
+  def self.parse_json(response)
+    JSON.parse(response.body, symbolize_names: true)
+  end
+end
--- a/app/views/foods/index.html.erb
+++ b/app/views/foods/index.html.erb
@@ -1,15 +1,11 @@
 <h3>Seach Results for <%= @foods[:foodSearchCriteria][:query] %></h3>
 <p>Total matching results: <%= @foods[:totalHits] %></p>
 <p>Ten foods that contain <%= @foods[:foodSearchCriteria][:query] %> </p>
-  <% @foods[:foods].each do |food| %>
-  <br>
-    <strong> Food </strong>
-    GTIN/UPC Code: <%= food[:gtinUpc] %>
-    <br>
-    Description: <%= food[:description] %>
-    <br>
-    Brand Owner: <%= food[:brandOwner] %>
-    <br>
-    Ingredients: <%= food[:ingredients] %>
-    <br>
+  <% @foods[:foods].first(10).each do |food| %>
+    <ul class="food"> Food
+      <li class="GTIN/UPC"> GTIN/UPC Code: <%= food[:gtinUpc] %> </li>
+      <li class="Description"> Description: <%= food[:description] %> </li>
+      <li class="Brand Owner"> Brand Owner: <%= food[:brandOwner] %> </li>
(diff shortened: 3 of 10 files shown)


app/facades/foods_facade.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def self.search_food_query(query)
+    data = FoodsService.search_food(query)
+    data[:foodSearchCriteria][:query]
+
+  def self.search_food_total(query)
+    data = FoodsService.search_food(query)
+    data[:totalHits]
+    end
```
timomitchel, line 17: This would be a great place to use memoization, because you only need to hit the API once with one call to the service. Then, if you had isolated that method, you could use memoization to either hit the API if it was the first time the method runs, or simply return the populated hash if the method had already run once. I'm thinking something like: 
```ruby
def fetch_food_data(query)
data ||= FoodsService.search_food(query)
end
```

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,19 @@
+class FoodsFacade
+  def self.search_food(query)
+    foods = []
+    data = FoodsService.search_food(query)
+    data[:foods].first(10).each do |food|
```
timomitchel, line 5: This is exactly where I would like to see the JSON scoped down to the `first(10)` objects. This way we only ever create the objects the view needs, instead of thousands of objects.

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+  end
+
+  def self.search_food_query(query)
+    data = FoodsService.search_food(query)
+    data[:foodSearchCriteria][:query]
+
+  def self.search_food_total(query)
+    data = FoodsService.search_food(query)
+    data[:totalHits]
```
timomitchel, line 16: Happy to see that you found this `totalHits` key.

app/services/foods_service.rb
```diff
@@ -0,0 +1,22 @@
+class FoodsService
+
+  def self.conn
+      Faraday.new(
+      url: "https://api.nal.usda.gov/fdc/",
+      params: {},
+      headers: {"X-Api-Key" => "#{ENV["FOOD_DATA_API_KEY"]}"
+    }
```
timomitchel, line 8: Nice job setting this on new lines as arguments. This is readable/maintainable.

app/services/foods_service.rb
```diff
@@ -0,0 +1,22 @@
...
+    )
+  end
+
+  def self.search_food(query)
+    response = conn.get("v1/foods/search?query=#{query}")
```
timomitchel, line 13: Would have loved to see the `query` param set in a block or as an argument similar to `conn` above.

app/services/foods_service.rb
```diff
@@ -0,0 +1,22 @@
...
+    parse_json(response)
+  end
+
+  private
+
+  def self.parse_json(response)
```
timomitchel, line 19: `conn` should also be a private method.

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,11 @@
...
+<p>Total matching results: <%= @food_total_results %></p>
+<p>Ten foods that contain <%= @food_title  %> </p>
+  <% @foods.each do |food| %>
+    <ul class="food"> Food
+      <li class="GTIN/UPC"> GTIN/UPC Code: <%= food.upc %> </li>
+      <li class="Description"> Description: <%= food.description %> </li>
+      <li class="Brand Owner"> Brand Owner: <%= food.brand_owner %> </li>
+      <li class="Ingredients"> Ingredients: <%= food.ingredients %> </li>
+  </ul>
+  <% end %>
```
timomitchel, line 11: Nice job here. Love to see you set the `class` for some of the elements.

config/routes.rb
```diff
@@ -1,3 +1,5 @@
 Rails.application.routes.draw do
   root "welcome#index"
+
+  resources :foods, only: [:index]
```
timomitchel, line 4: Nice job only creating the one route we need.

spec/features/search_food_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      expect(current_path).to eq(foods_path)
+      expect(page.status_code).to eq(200)
+      expect(page.all('ul', count: 10))
+
+      expect(page).to have_content("Total matching results")
+      expect(page).to have_content("Description")
+      expect(page).to have_content("Brand Owner")
+      expect(page).to have_content("Ingredients")
+      expect(page).to have_content("sweet potatoes")
```
timomitchel, line 34: Solid testing here. Might be nice to also check for the first and last Food object data we expect to see on the page.

spec/services/foods_service_spec.rb
```diff
@@ -0,0 +1,11 @@
+require "rails_helper"
+
+RSpec.describe FoodsService do
+  it "returns correct structure for search foods method" do
+    response = FoodsService.search_food("sweet potatoes")
+    expect(response).to be_a(Hash)
+    expect(response[:foods]).to be_a(Array)
+    expect(response[:foods][1]).to have_key(:gtinUpc)
+    expect(response[:results][1][:gtinUPC]).to be_a(String)
```
timomitchel, line 9: Nice, this looks to be on its way and is testing the structure of our JSON response nicely.

timomitchel (commented):
Overall, a good job on this challenge Keegan. You're showing a ton of progress. LMK if you have questions on the feedback.

PR #25, opened 2021-05-24
Austin Andrade

austinandrade:
- Complete all base requirements including testing
- Miss implementation of Webmock/VCR

### DIFF
--- a/app/facades/foods_facade.rb
+++ b/app/facades/foods_facade.rb
@@ -1,0 +1,16 @@
+class FoodsFacade
+  class << self
+    def relevant_dishes(ingredient)
+      create_dishes(FoodsService.get_dishes_including_ingredient(ingredient))
+    end
+
+    private
+
+    def create_dishes(dish_data)
+      total_found = dish_data[:totalHits]
+      dish_data[:foods].map do |dishy_data|
+        Food.new(dishy_data, total_found)
+      end
+    end
+  end
+end
--- a/app/services/foods_service.rb
+++ b/app/services/foods_service.rb
@@ -1,0 +1,20 @@
+class FoodsService
+  class << self
+    def get_dishes_including_ingredient(ingredient)
+      response = conn.get("search/?query=#{ingredient}&pageSize=10")
+      parse_data(response)
+    end
+
+    private
+
+    def conn
+      Faraday.new(url: 'https://api.nal.usda.gov/fdc/v1/') do |faraday|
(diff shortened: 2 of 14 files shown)


app/facades/foods_facade.rb
```diff
@@ -0,0 +1,16 @@
...
+      create_dishes(FoodsService.get_dishes_including_ingredient(ingredient))
+    end
+
+    private
+
+    def create_dishes(dish_data)
+      total_found = dish_data[:totalHits]
+      dish_data[:foods].map do |dishy_data|
+        Food.new(dishy_data, total_found)
+      end
```
megstang, line 13: this `total_found` doesn't really have much to do with the individual food objects, but it does have something to do with the original query. We definitely do want to get this sent back, and I like that you're trying to wrap it in with the objects you're already creating. Something you could look into is the following syntax: 

`total_found, foods= FoodFacade.your_method(params[:q])` 

You could put this in your controller, and then have `your_method` return a 2d array where the first element was the totalHits, and the second element was the collection of food objects you created.

app/services/foods_service.rb
```diff
@@ -0,0 +1,20 @@
+class FoodsService
+  class << self
+    def get_dishes_including_ingredient(ingredient)
+      response = conn.get("search/?query=#{ingredient}&pageSize=10")
+      parse_data(response)
+    end
```
megstang, line 6: good call using the pageSize param here

megstang (commented):
@austinandrade Nice work, I left just a few comments. Take a look over them and let me know if you have any questions. I'll be following up soon with scores via slack.

app/facades/foods_facade.rb
megstang: My thought around suggesting this is to get you thinking about a better way to format the two pieces of data we need to display to the view. The totalHits value is the total number of hits you have from the query that was made, and currently, you're saving that number of total hits to EVERY food you create -- it's the same number, and it's being stored on every Food object. Instead of doing that, you could send it back to your controller. 

```ruby
     total_found = dish_data[:totalHits]

     foods_collection = dish_data[:foods].map do |dishy_data|        
         Food.new(dishy_data)      
      end

     [total_found, foods_collection]
```

megstang: When you return an array, you can assign it to variables by using the syntax I put in my first comment above.

PR #26, opened 2021-05-24
Logan

loganjacob76:
This mid mod was fine for the most part, but unfortunately I spent over a quarter of my time troubleshooting my service because it was returning an undefined error method. Had that not been the case I would have cleaned things up a bit more, by doing things like implementing vcr perhaps and I would have added poro tests.

### DIFF
--- a/app/controllers/foods_controller.rb
+++ b/app/controllers/foods_controller.rb
@@ -1,0 +1,5 @@
+class FoodsController < ApplicationController
+  def index
+    # binding.pry
+  end
+end
--- a/app/controllers/foods_controller.rb
+++ b/app/controllers/foods_controller.rb
@@ -1,5 +1,5 @@
 class FoodsController < ApplicationController
   def index
-    # binding.pry
+    @results = FoodFacade.foods_by_ingredient(params[:q])
   end
 end
--- a/app/facades/food_facade.rb
+++ b/app/facades/food_facade.rb
@@ -2,8 +2,7 @@ class FoodFacade
   class << self
     def foods_by_ingredient(ingredient)
       results = FoodService.search_by_ingredient(ingredient)
-      results[:foods] = results[:foods][0..9]
-      results
+      Search.new(results)
     end
   end
 end
--- a/app/facades/food_facade.rb
+++ b/app/facades/food_facade.rb
@@ -1,0 +1,9 @@
+class FoodFacade
+  class << self
+    def foods_by_ingredient(ingredient)
+      results = FoodService.search_by_ingredient(ingredient)
+      results[:foods] = results[:foods][0..9]
+      results
+    end
+  end
+end
--- a/app/poros/food.rb
+++ b/app/poros/food.rb
@@ -1,0 +1,13 @@
+class Food
+  attr_reader :brand_owner,
+              :gtin_upc,
+              :ingredients,
+              :description
+
+  def initialize(food)
+    @brand_owner = food[:brandOwner]
+    @gtin_upc = food[:gtinUpc]
+    @ingredients = food[:ingredients].downcase
+    @description = food[:lowercaseDescription]
+  end
+end
--- a/app/poros/search.rb
+++ b/app/poros/search.rb
@@ -1,0 +1,16 @@
+class Search
+  attr_reader :total_results,
+              :query
+  
+  def initialize(results)
+    @total_results = results[:totalHits]
+    @query = results[:foodSearchCriteria][:query].capitalize
+    @all_foods = results[:foods]
+  end
+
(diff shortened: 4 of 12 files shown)


app/poros/search.rb
```diff
@@ -0,0 +1,16 @@
+class Search
```
timomitchel, line 1: I like that you made one presentation object to pass to the view. This is a good pattern.

app/poros/food.rb
```diff
@@ -0,0 +1,13 @@
+class Food
+  attr_reader :brand_owner,
```
timomitchel, line 2: Solid PORO here.

app/facades/food_facade.rb
```diff
@@ -0,0 +1,8 @@
+class FoodFacade
+  class << self
+    def foods_by_ingredient(ingredient)
+      results = FoodService.search_by_ingredient(ingredient)
+      Search.new(results)
```
timomitchel, line 5: Essentially, your `Search` object is your Presenter. It's a good pattern that we haven't taught.

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,5 @@
+class FoodsController < ApplicationController
+  def index
+    @results = FoodFacade.foods_by_ingredient(params[:q])
```
timomitchel, line 3: Love that this is one line, but I might call the instance variable `@search` to match the name of your object.

app/poros/search.rb
```diff
@@ -0,0 +1,16 @@
...
+              :query
+  
+  def initialize(results)
+    @total_results = results[:totalHits]
+    @query = results[:foodSearchCriteria][:query].capitalize
+    @all_foods = results[:foods]
+  end
+
+  def foods
+    @all_foods[0..9].map do |food|
```
timomitchel, line 12: I'm glad you only created 10 Food object. This is efficient. And, this is the right place to do this.

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,14 @@
...
+<ol>
+  <% @results.foods.each do |food| %>
+    <li>Brand Owner: <%= food.brand_owner %></li>
+    <ul>
+      <li>GTIN/UPC: <%= food.gtin_upc %></li>
+      <li>Ingredients: <%= food.ingredients %></li>
+      <li>Description: <%= food.description %></li>
+    </ul> <br>
+  <% end %>
+</ol>
```
timomitchel, line 14: Clean and effective.

config/routes.rb
```diff
@@ -1,3 +1,4 @@
 Rails.application.routes.draw do
   root "welcome#index"
+  resources :foods
```
timomitchel, line 3: Would have liked to see the `only` syntax used in conjunction with this view, so we aren't creating unused routes.

spec/facades/food_facade_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'rails_helper'
+
+RSpec.describe FoodFacade do
+  it 'formats all desired info' do
+    results = FoodFacade.foods_by_ingredient('sweet potatoes')
+
+    expect(results.total_results).to eq(44128)
+    expect(results.query).to eq('Sweet potatoes')
+    expect(results.foods.count).to eq(10)
```
timomitchel, line 9: Solid. I might also check that `results.foods` is an `Array` and the first object in the array is a `Food` object. Also, would be good to stub out the Service API call that this facade method depends on .

app/services/food_service.rb
```diff
@@ -0,0 +1,18 @@
...
+      end
+      JSON.parse(resp.body, symbolize_names: true)
+    end
+    
+    private
+    
+    def conn
+      Faraday.new(url: 'https://api.nal.usda.gov/fdc/v1/') do |faraday|
+        faraday.params['api_key'] = ENV['food_key']
+      end
```
timomitchel, line 15: This class looks great.

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+      expect(current_path).to eq(foods_path)
+    end
+  end
+  
+  context 'you have searched sweet potatoes' do
+    before :each do
+      visit root_path
+      fill_in :q, with: 'sweet potatoes'
+      click_on :commit
+    end
```
timomitchel, line 19: These could be combined. Really the top it block should be within the context block you begin on 14. Again, the API call should be stubbed out.

timomitchel (commented):
Overall, a great performance on this mid mod. LMK if you have questions on the feedback.

PR #27, opened 2021-05-24
Alexander Brueck

brueck1988:
I think the midmod went really well.


spec/poros/food_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+      expect(food.gtinupc).to eq("492111402857")
+
+      expect(food.description.class).to eq(String)
+      expect(food.description).to eq("SWEET POTATOES")
+
+      expect(food.brandowner.class).to eq(String)
+      expect(food.brandowner).to eq("ARCHER FARMS")
+
+      expect(food.ingredients.class).to eq(String)
+      expect(food.ingredients).to eq("SWEET POTATOES.")
```
megstang, line 160: I don't think you really need to test the class of these attrs since you're checking what they're equal to anyway. But this is a super thorough test.. nice work.

app/services/food_service.rb
```diff
@@ -0,0 +1,10 @@
+class FoodService
+  def self.find_foods_by_ingredient(food)
+    response = Faraday.get("https://api.nal.usda.gov/fdc/v1/foods/search?") do |faraday|
+      faraday.params[:query] = "#{food}"
+      faraday.params[:pageSize] = 10
+      faraday.params[:api_key] = "#{ENV['api_key']}"
+    end
+    parse = JSON.parse(response.body, symbolize_names: true)
+  end
+end
```
megstang, line 10: This all looks good to me -- I know we talk quite a bit about abstracting the connection, but in the case of this mid mod, when we've asked for nothing other than this one user story, there wasn't really a need for that. As soon as you needed to make another call to that api, that's when I'd make sure to abstract the connection into it's own method. I also like how you're including the pageSize in the query -- nice job reading the docs.

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,6 @@
+class FoodsController < ApplicationController
+  def index
+    @foods = FoodFacade.find_foods_by_ingredient(params[:q])
+    @number_of_results = FoodFacade.find_number_of_results(params[:q])
+  end
```
megstang, line 5: Both of these methods, `find_foods_by_ingredient` and `find_number_of_results` call the same endpoint, and just pull different data. I'd like to challenge you to find a way to just call one method in your facade, and have your facade take care of combining the food objects and the totalHits into one object to get returned to your controller. Ideally, you just have one object getting sent to your view that contains all data necessary for that view. And in this case, all of the data that you need for this view, can come from the one query call you're making.

megstang: You could take a look at this syntax: 

`number_of_results, foods = FoodFacade.your_method(params[:q])`

Where `your_method` returns a 2d array, with the number of results as the first element and the food collection as the second element.

megstang (commented):
@brueck1988 Great job! I left just a few comments. Please let me know if you have any questions. I'll DM you shortly with your scores.

PR #32, opened 2021-10-25
Tanner Dale

TannerDale:


app/clients/food_client.rb
```diff
@@ -0,0 +1,18 @@
...
+
+    def conn
+      Faraday.new('https://api.nal.usda.gov')
+    end
+
+    def parse_data(response)
+      JSON.parse(response.body, symbolize_names: true)
+    end
+  end
+end
```
epintozzi, line 18: Often when I see a client used, it's for the initial setup for making your requests (ex: faraday, the base url and any params or headers we always send). Then, there'd be a Service that calls the client for our specific endpoint or project needs. That way we can have many Services with one Client. Because this is such a small project, that would be overkill here. I'm only including this info a kind of a side note since I don't believe we actually teach Client. It is a class I often see for API consumption so I don't mind that you're using it over a Service. This is truly just an extra piece of info and not a request to do it differently, so long as you understand what you are doing.

app/facades/food_facade.rb
```diff
@@ -0,0 +1,17 @@
+class FoodFacade
+  class << self
+    def search(name)
+      data = FoodClient.search(name)
+
+      [foods(data), data[:totalHits]]
```
epintozzi, line 6: I really like how you're handling this and not making two separate API calls to the same endpoint for these two pieces of information. Nice job.

app/clients/food_client.rb
```diff
@@ -0,0 +1,18 @@
+class FoodClient
+  class << self
+    def search(food)
+      url = "/fdc/v1/foods/search?query=#{food}&pageSize=10&api_key=#{ENV['api_key']}"
```
epintozzi, line 4: Nice job catching the `pageSize` param option.

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      FoodPoro.new({ gtinUpc: 456, lowercaseDescription: 'yellow', brandOwner: 'mito', ingredients: 'peas, elephant' }),
+      FoodPoro.new({ gtinUpc: 789, lowercaseDescription: 'orange', brandOwner: nil, ingredients: 'hippo, broccoli' })
+    ]
+  end
+
+  before :each do
+    allow(FoodFacade).to receive(:search).and_return([foods, 12345])
+
+    visit foods_path
+  end
```
epintozzi, line 16: Good use of stubbing here. It allows you the flexibility to test the specifics that you expect to see in the view without getting caught up in whether or not the API is working correctly (since you are presumably testing that elsewhere)

epintozzi:
Nice job! I'll send over some additional feedback via slack shortly. Let me know if you have questions.

PR #33, opened 2021-10-25
Ezzedine Alwafai

ealwafai:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,7 @@
+class FoodsController < ApplicationController
+  def index
+    results = FoodFacade.search_by_ingredient(params[:q])
+    @result_count = results[:total_result_count]
+    @foods = results[:top_ten_foods]
+  end
```
megstang, line 6: Nice job here -- I like that you're returning both values from the `search_by_ingredient` method. Since you already have those two values wrapped in one object, you could just send that one object to the view instead of creating two. If you chose to do that, I'd rename `results` on line 3 to `food_results` and then get the values at the specific keys in the view.

app/services/food_service.rb
```diff
@@ -0,0 +1,11 @@
+class FoodService
+
+  def self.search_by_ingredient(ingredient)
+    response = Faraday.get('https://api.nal.usda.gov/fdc/v1/foods/search') do |req|
+      req.params['api_key'] = ENV['food_key']
+      req.params['query'] = ingredient
+      req.params['pageSize'] = 10
+    end
+    JSON.parse(response.body, symbolize_names: true)
+  end
```
megstang, line 10: great use of pageSize here to limit the return value.

app/poros/food.rb
```diff
@@ -0,0 +1,10 @@
+class Food
+  attr_reader :upc_code, :description, :brand, :ingredients
+
+  def initialize(food_info)
+    @upc_code = food_info[:gtinUpc]
+    @description = food_info[:description]
+    @brand = food_info[:brandOwner]
+    @ingredients = food_info[:ingredients]
+  end
+end
```

app/facades/food_facade.rb
```diff
@@ -0,0 +1,11 @@
...
+  def self.search_by_ingredient(ingredient)
+    result = FoodService.search_by_ingredient(ingredient)
+    {
+      total_result_count: result[:totalHits],
+      top_ten_foods: result[:foods].map do |food|
+        Food.new(food)
+      end
+    }
+  end
+end
```
megstang, line 11: 👍

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,10 @@
+<h1>Search Results</h1>
+
+<h3>Total Foods: <%= @result_count %></h3>
```
megstang, line 3: If you went with just sending one object to the view like I mentioned above, you could just do: 
```
<h3>Total Foods: <%= @food_results[:total_result_count] %></h3>
```

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,10 @@
...
+
+<% @foods.each do |food| %>
```
megstang, line 5: similarly here... 

```
<% @food_results[:top_ten_foods].each do |food| %>
```

spec/facades/food_facade_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+      expect(results).to have_key(:total_result_count)
+      expect(results[:total_result_count]).to eq(46320)
+      expect(results).to have_key(:top_ten_foods)
+      expect(results[:top_ten_foods]).to be_an(Array)
+
+      expect(results[:top_ten_foods].first).to be_a(Food)
+      expect(results[:top_ten_foods].first.upc_code).to eq('076700002019')
+      expect(results[:top_ten_foods].first.description).to eq('SWEET POTATOES')
+      expect(results[:top_ten_foods].first.brand).to eq('John W. Taylor Packing Co. Inc')
+      expect(results[:top_ten_foods].first.ingredients).to eq('SWEET POTATOES.')
```
megstang, line 18: If you're testing your food poro, you don't need to have these tests.

megstang: Looks like you don't have a test for your food poro -- that spec can just test that a food poro can get created and that it has the attributes you'd expect a food object to have.

megstang (commented):
@ealwafai Great job! Look over these comments and let me know if you have any questions. I'll be following up via slack before the end of the day with more feedback.

PR #34, opened 2021-10-25
Weston Ellis

WMudgeEllis:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,7 @@
+class FoodsController < ApplicationController
+
+  def search
+    @results = FoodFacade.food_search(params[:q])
```
megstang, line 4: Really nice work getting this method to return both the totalHits and the food objects 👍

app/facades/food_facade.rb
```diff
@@ -0,0 +1,9 @@
+class FoodFacade
+
+  def self.food_search(query)
+    results = FoodService.search(query)
+    foods = results[:foods][0..9].map { |food_data| Food.new(food_data) }
+    data = { total_results: results[:totalHits], foods: foods}
+    SearchResult.new(data)
```
megstang, line 7: 🎉 Love this idea of making a new object for combining these results.

app/facades/food_facade.rb
```diff
@@ -0,0 +1,9 @@
+class FoodFacade
+
+  def self.food_search(query)
+    results = FoodService.search(query)
+    foods = results[:foods][0..9].map { |food_data| Food.new(food_data) }
```
megstang, line 5: You may consider using the perPage parameter to limit how many foods you get back with just one call. That parameter is unique to this API, but it looks like it's something that can be used.

app/poros/food.rb
```diff
@@ -0,0 +1,10 @@
+class Food
+  attr_reader :id, :description, :owner, :ingredients
+
+  def initialize(data)
+    @id = data[:gtinUpc]
+    @description = data[:description]
+    @owner = data[:brandOwner]
+    @ingredients = data[:ingredients]
+  end
+end
```

app/poros/search_result.rb
```diff
@@ -0,0 +1,8 @@
+class SearchResult
+  attr_reader :total_results, :foods
+
+  def initialize(data)
+    @total_results = data[:total_results]
+    @foods = data[:foods]
+  end
```
megstang, line 7: 👍

app/services/food_service.rb
```diff
@@ -0,0 +1,14 @@
+class FoodService
+
+
+  def self.search(query)
+    response = Faraday.get('https://api.nal.usda.gov/fdc/v1/foods/search') do |f|
+      f.params['api_key'] = ENV['api_key']
+      f.params['query'] = query
+      f.params['dataType'] = 'Branded'
+    end
```
megstang, line 9: You mentioned you weren't sure if you were using the right endpoint or not -- you are. You must have seen a little edge case, that not all foods that were returned from this search are branded, which then leads the food responses to not have some of the values we're hoping to display. So, nice catch on that edge case!

megstang: I like that you're using the block to assign parameters.

spec/services/food_service_spec.rb
```diff
@@ -0,0 +1,12 @@
+require "rails_helper"
+
+RSpec.describe FoodService do
+  it 'can get search info' do
+    results = FoodService.search('sweet potatoes')
+
+    expect(results).to be_a(Hash)
+    expect(results[:totalHits]).to eq(45082)
+    expect(results).to have_key(:foods)
+    expect(results[:foods]).to be_a(Array)
```
megstang, line 10: I would also jump into that first foods value and test the keys that you'd expect to be associated with any food (description, ingredients, etc.)

megstang (commented):
@WMudgeEllis Really awesome job. I'll follow up via slack dm shortly with some more feedback, but please let me know if you have any questions on any of my feedback.

PR #35, opened 2021-10-25
Henry Schmid

hschmid516:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,6 @@
+class FoodsController < ApplicationController
+  def index
+    @search_query = params[:q]
+    @foods = FoodFacade.find_foods('sweet potatoes')
```
epintozzi, line 4: I know you already called this out but for the sake of being thorough, we should be passing the actual value of the query param here 😂 . I pulled your project and it does indeed work as expected when I do so.

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,13 @@
+<h1>Search Results for "<%= @search_query %>"</h1>
+
+<p>Total Results: <%= @foods.first.total_hits %></p>
+
+<h4>Top 10 Results</h4>
+<% @foods[0..9].each do |food| %>
```
epintozzi, line 6: `[0..9]` works to accomplish what we are trying to do. There's also a param we can pass (`pageSize`) when making our API call to limit the number of results returned. That would be more efficient than holding 40 extra poros in memory. It will also give you a fast response when making the call as well as less strain on the service providing the data.

spec/poros/food_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'rails_helper'
+
+describe Food do
+  it 'has attributes' do
+    food = Food.new(food_info, total_hits)
+  end
```
epintozzi, line 6: We want to make sure we're actually testing something here. Right now there are no assertions. We could do something like:

`expect(food.gtin_upc).to eq('123456')` for each attribute (inserting the correct values of course)

spec/facades/food_facade_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'rails_helper'
+
+describe FoodFacade do
+  it 'can find foods' do
+    foods = FoodFacade.find_foods('sweet potatoes')
+
+    expect(foods.first).to be_a Food
+    expect(foods.first.id).to eq(1929370)
```
epintozzi, line 8: This is probably better suited to be testing in the `Food` poro. In addition to testing our Facade returns a `Food` object, we could also test that it's an array and how many items are in that array.

epintozzi:
Overall great job on the midmod. Just some super minor comments. I'll send more feedback via slack before the end of the day. Let me know if you have any questions!

PR #36, opened 2021-10-25
Jamie Pace

jamiejpace:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,6 @@
+class FoodsController < ApplicationController
+  def index
+    @foods = FoodFacade.food_search(params[:q])
+    @results_count = FoodFacade.results_count(params[:q])
```
megstang, line 4: Both of these facade methods, `food_search` and `results_count` are eventually calling to the same service method, which means it's calling to the same endpoint. Since both totalHits AND food data is returned by calling to that endpoint once, let's consider an approach that would pull out the data you need for your view with just ONE api call: 

```ruby
@foods, @results_count = FoodFacade.food_search(params[:q])
```

So, in your `food_search` action in your facade, you could do something like: 

```ruby
def food_search(food)
      foods = data[:foods].take(10).map do |food_data|
        Food.new(food_data)
      end
      [foods, data[:totalHits]]
    end
```

You could also consider sending back a hash like this: 

```ruby
def food_search(food)
      foods = data[:foods].take(10).map do |food_data|
        Food.new(food_data)
      end
      {foods: foods, total_hits: data[:totalHits]}
    end
```

and then you could send that hash to the view, and have your view grab the values it needs when it wants to display something.

app/poros/food.rb
```diff
@@ -0,0 +1,11 @@
+class Food
+  attr_reader :code, :description, :brand_owner, :food_ingredients, :name
+
+  def initialize(data)
+    @name = data[:brandName]
+    @code = data[:gtinUpc]
+    @description = data[:description]
+    @brand_owner = data[:brandOwner]
+    @food_ingredients = data[:ingredients]
+  end
```

app/services/food_service.rb
```diff
@@ -0,0 +1,20 @@
+class FoodService
+  class << self
+    def find_foods(food)
+      response = conn.get("/fdc/v1/foods/search?query=#{food}")
```
megstang, line 4: You can also assign parameters in a block just like you did in the conn method for the api_key. 
```ruby
conn.get("/fdc/v1/foods/search") do |req|
   req.params['query'] = food
end 
```

megstang: ^^ This way is more conventional

spec/facades/food_facade_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'rails_helper'
+
+RSpec.describe 'food facade', :vcr do
+  it 'can return 10 food objects' do
+    foods = FoodFacade.food_search("chocolate chips")
+
+    expect(foods).to be_an(Array)
+    expect(foods.first).to be_a(Food)
+    expect(foods.first.description).to be_a(String)
```
megstang, line 9: You don't need to check the attributes on the food poro, since your food poro spec should be testing the attributes. This is an example of how overtesting can happen.

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  it 'displays a list of foods from search' do
+    visit root_path
+
+    fill_in :q, with: "sweet potatoes"
+    click_on "Search"
+
+    expect(current_path).to eq(foods_path)
+    expect(page).to have_content("Brand Owner", count: 10)
+    expect(page).to have_content("GTIN/UPC Code:", count: 10)
+    expect(page).to have_content("Ingredients:", count: 10)
```
megstang, line 13: Nice call here, counting the times it occurs on the page 👍

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+    expect(page).to have_content("Search Results")
+    expect(page).to have_content("Total Number of Results:")
+    expect(page).to have_content("Top 10 Matches:")
+  end
```
megstang, line 17: This is all good -- you could build this out a bit by testing the values on the page. Since your facade method creates food objects for you, and your facade method is already tested, you could call your facade method in here that returns the food objects, and test to make sure that the data that comes back from those food objects, is occuring on this page. I don't think it's necessary to test for all 10 foods and their data, BUT it could be a bit more thorough if you found a way to make sure that the attributes were getting displayed on the page the way you'd expect them too, because right now, you're just checking that the labels are there for the attributes.

spec/services/food_service_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'rails_helper'
+
+RSpec.describe 'food service', :vcr do
+  it 'returns a list of foods from a search' do
+    response = FoodService.find_foods("chocolate chips")
+
+    expect(response).to be_a(Hash)
+    expect(response).to have_key(:totalHits)
+    expect(response).to have_key(:foods)
```
megstang, line 9: Also, go into the foods hash and check for the keys that are in there that you'll want to use (brandOwner, ingredients, description, etc.)

megstang: `expect(response[:foods]).to have_key(:brandOwner)` ... and so on...

megstang (commented):
@jamiejpace Nice job! I left a few comments for you to look over. Please let me know if you have any questions. I'll follow up via slack with some more feedback before the end of the day.

PR #37, opened 2021-10-25
Jacq Whitmer

jrwhitmer:


config/routes.rb
```diff
@@ -1,3 +1,4 @@
 Rails.application.routes.draw do
   root "welcome#index"
+  resources :foods
```
epintozzi, line 3: Since we are only using the index we can write this instead as `resources :foods, only: :index`

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      expect(page).to have_content("819614010394")
+    end
+  end
+
+  it 'shows 10 food results on the page' do
+    VCR.use_cassette('search-1') do
+      @foods = FoodsFacade.search('sweet potato')
+
+      expect(@foods.length).to eq(10)
+    end
```
epintozzi, line 45: I like that you are explicitly testing everything we'd like to see here. Because you have the `before` action, we are going through the motions of submitting the form for each spec. Depending on what your future work team looks like, it might be preferred to do that just once and make all your assertions in one `it` block. On the flip side, it can also be nice to have a dedicated spec name for each thing we are testing for like you have here. I personally don't have strong feelings either way, particularly because you are using vcr and therefore we aren't waiting for real API calls here, but just a note for other ways you might see this done.

spec/rails_helper.rb
```diff
@@ -61,4 +61,9 @@
   config.filter_rails_from_backtrace!
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
+
+  VCR.configure do |config|
+    config.cassette_library_dir = "fixtures/vcr_cassettes"
+    config.hook_into :webmock
+  end
```
epintozzi, line 68: Since you are committing and pushing your cassettes to github you should make sure you are including the config to remove your api key from your cassettes with `config.filter_sensitive_data('<food_key>') { ENV['API_KEY'] }`. Right now all your cassettes have you key in them which is not a great thing to have up on github. I'd strongly recommend going back and cleaning it up.

app/services/foods_service.rb
```diff
@@ -0,0 +1,17 @@
+class FoodsService
+  class << self
+    def get_data(endpoint)
+      response = Faraday.get(endpoint)
+      JSON.parse(response.body, symbolize_names: true)
+    end
+
+    def food_search(keyword)
+      foods = get_data("https://api.nal.usda.gov/fdc/v1/foods/search?query=#{keyword}&pageSize=10&api_key=#{ENV['API_KEY']}")
```
epintozzi, line 9: I really like that you are limiting to 10 results here rather than returning everything and limiting later 👍🏻

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,10 @@
+class FoodsController < ApplicationController
+
+  def index
+    @foods = if params[:q].present?
+      FoodsFacade.search(params[:q])
+    end
+    @search_info = FoodsFacade.search_hits(params[:q])
```
epintozzi, line 7: `search` and `search_hits` appear to call identical methods in the service (as `food_search` and `food_search_hits` so we don't really need to have them separated and call them twice. You can handle the search hits data from the call you make for `search`.

epintozzi:
Nice job! I'll send more feeback via slack later today. Let me know if you have any questions.

PR #38, opened 2021-10-25
Kelsey Thompson

knthompson2:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,6 @@
+class FoodsController < ApplicationController
+  def index
+    @food_count = FoodFacade.total_count(params[:q])
+    @foods = FoodFacade.search_by_food(params[:q])
```
megstang, line 4: Both of these facade methods, `total_count` and `search_by_food` are eventually calling to the same service method, which means it's calling to the same endpoint. Since both totalHits AND food data is returned by calling to that endpoint once, let's consider an approach that would pull out the data you need for your view with just ONE api call: 

```ruby
@foods, @food_count = FoodFacade.search_by_food(params[:q])
```

So, in your `search_by_food` action in your facade, you could do something like: 

```ruby
def search_by_food(food)
      data = FoodService.food_search(food)[:foods][0..9]
    foods = data.map do |food|
      Food.new(food)
    end
      [foods, data[:totalHits]]
    end
```

You could also consider sending back a hash like this: 

```ruby
def search_by_food(food)
      data = FoodService.food_search(food)[:foods][0..9]
    foods = data.map do |food|
      Food.new(food)
    end
      {foods: foods, total_hits: data[:totalHits]}
    end
```

and then you could send that hash to the view, and have your view grab the values it needs when it wants to display something.

app/facades/food_facade.rb
```diff
@@ -0,0 +1,12 @@
+class FoodFacade
+  def self.search_by_food(food)
+    foods = FoodService.food_search(food)[:foods][0..9]
```
megstang, line 3: instead of limiting the results here, you could add the perPage parameter onto the request to get back 10 foods right away.

app/poros/food.rb
```diff
@@ -0,0 +1,13 @@
...
+              :brand_owner,
+              :ingredients
+
+  def initialize(info)
+    @upc_code = info[:fdcId]
+    @description = info[:description]
+    @brand_owner = info[:brandOwner]
+    @ingredients = info[:ingredients]
+  end
+end
```
megstang, line 13: 👍

app/services/food_service.rb
```diff
@@ -0,0 +1,19 @@
...
+  def self.conn
+    Faraday.new('https://api.nal.usda.gov/fdc/') do |f|
+      f.params['api_key'] = ENV['food_key']
+    end
+  end
+
+  def self.parse_json(response)
+    JSON.parse(response.body, symbolize_names: true)
+  end
+end
```
megstang, line 19: 👍

spec/facades/food_facade_spec.rb
```diff
@@ -0,0 +1,12 @@
...
+RSpec.describe FoodFacade do
+  describe '.search_by_food' do
+    it 'returns a list of foods by search query' do
+      search_results = FoodFacade.search_by_food("sweet potatoes")
+
+      expect(search_results).to be_a(Array)
+      expect(search_results.count).to eq(10)
+    end
+  end
+end
```
megstang, line 12: I would also check in here that the elements that are returned in that array are Food objects.

spec/features/index_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    expect(page).to have_content(total)
+    foods.each do |food|
+      expect(page).to have_content(food.upc_code)
+      expect(page).to have_content(food.description)
+      expect(page).to have_content(food.brand_owner)
+
+        expect(page).to have_content(food.ingredients)
+    end
+  end
+end
```
megstang, line 21: 👍

spec/facades/food_facade_spec.rb
megstang: Also, add a spec to test your poro -- in that food_poro_spec, you can test that a food object can get created and that it can read the attributes you'd want to be able to store on the object.

megstang (commented):
@knthompson2 Really great job. I left a few comments for you to look over, please let me know if you have any questions. I'll follow up via slack dm with some more feedback before the end of the day.

PR #39, opened 2021-10-25
Ida Olson

idaolson:


app/facades/food_facade.rb
```diff
@@ -0,0 +1,9 @@
+class FoodFacade
+  def self.search(query)
+    data = FdcService.search(query)
+    foods = data[:foods].map do |food|
+      Food.new(food)
+    end
+    {total_hits: data[:totalHits], top_ten: foods}
+  end
```
megstang, line 8: 👍

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,7 @@
+class FoodsController < ApplicationController
+  def index
+    results = FoodFacade.search(params[:q])
+    @total_hits = results[:total_hits]
+    @top_ten = results[:top_ten]
```
megstang, line 5: This looks great! Nice job finding a way to get both the totalHits AND the foods collection from just one call. You could also send the results directly to the view and have your view access the keys it needs. That way, you're only sending one object to the view.

app/poros/food.rb
```diff
@@ -0,0 +1,15 @@
...
+  attr_reader :upc, :description, :brand_owner, :ingredients
+
+  def initialize(params)
+    @upc = params[:gtinUpc]
+    @description = params[:description]
+    @brand_owner = params[:brandOwner]
+    @ingredients = params[:ingredients]
+  end
+end
+
```
megstang, line 11: 👍

app/services/fdc_service.rb
```diff
@@ -0,0 +1,6 @@
+class FdcService
+  def self.search(query)
+    response = Faraday.get("https://api.nal.usda.gov/fdc/v1/foods/search?query=#{query}&pageSize=10&api_key=#{ENV['food_api_key']}")
+    JSON.parse(response.body, symbolize_names: true)
+  end
```
megstang, line 5: I'd like to see you use some block variables for these faraday methods. Pass your parameters through by doing something like: 
```ruby
Faraday.get("https://api.nal.usda.gov/fdc/v1/foods/search) do |req|
    req.params[:api_key] = ENV['food_api_key']
    req.params[:query] = query 
     .......
end
```

megstang: also.. good call on using the perPage parameter to limit your response.

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,14 @@
...
+<% if @top_ten.count > 0 %>
+  <% @top_ten.each do |food| %>
+    <p>GTIN/UPC Code: <%= food.upc %></p>
+    <p>Description: <%= food.description %></p>
+    <p>Brand Owner: <%= food.brand_owner %></p>
+    <p>Ingredients: <%= food.ingredients %></p><br>
+  <% end %>
+<% else %>
+  <p>No results</p>
+<% end %>
```
megstang, line 14: 👍

spec/facades/food_facade_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  it 'returns total count and top 10 search results from query' do
+
+    expect(@results[:total_hits]).to eq(21874)
+    expect(@results[:top_ten].length).to eq(10)
+    expect(@results[:top_ten].first.upc).to eq("070038322238")
+    expect(@results[:top_ten].first.description).to eq("APPLE")
+    expect(@results[:top_ten].first.brand_owner).to eq("Associated Wholesale Grocers, Inc.")
+    expect(@results[:top_ten].first.ingredients).to eq("FILTERED WATER, APPLE JUICE CONCENTRATE, ASCORBIC ACID (VITAMIN C).")
```
megstang, line 16: If you're tesitng your food poro, there's no need to test these lines here, as the attributes for the food objects should be tested through the food poro spec. Here, you can just test that the top ten's elements are, in fact, food objects.

spec/features/welcome/index_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    expect(page).to have_button("Search")
+
+    fill_in :q, with: "sweet potatoes"
+    click_on "Search"
+
+    expect(current_path).to eq(foods_path)
+    expect(page).to have_content("GTIN/UPC Code: 076700002019")
+    expect(page).to have_content("Description: SWEET POTATOES")
+    expect(page).to have_content("Brand Owner: John W. Taylor Packing Co. Inc")
+    expect(page).to have_content("Ingredients: SWEET POTATOES.")
```
megstang, line 17: since you know there will be 10 foods returned, this may be a good test to implement a within block to check more than one food's data. You could also use css selectors in your test to make sure that there are 10 elements with a specific class.

spec/poros/food_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+              ingredients: "semolina"}
+
+    food = Food.new(params)
+
+    expect(food.upc).to eq(params[:gtinUpc])
+    expect(food.description).to eq(params[:description])
+    expect(food.brand_owner).to eq(params[:brandOwner])
+    expect(food.ingredients).to eq(params[:ingredients])
+  end
+end
```
megstang, line 18: 👍

megstang (commented):
@idaolson Great job! I left you some feedback to take a look over. Please let me know if you have any questions. I'll be following up via slack before the end of the day with some more feedback.

PR #40, opened 2021-10-25
Andrew Massey

acmassey3698:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,7 @@
+class FoodsController < ApplicationController
+  def index
+    @food_name  = params[:q]
+    @food_count = FoodsFacade.total_items(params[:q])
+    @ten_foods  = FoodsFacade.first_ten_foods(params[:q])
```
megstang, line 5: Both `total_items` and `first_ten_foods` call the same endpoint. Something you may consider doing is grabbing the datat you want from both of those methods and returning both at the same time. 

You could do something like: 
```ruby
@food_count, @ten_foods = FoodsFacade.food_search(params[:q])
```

and then your `food_search` method in the `FoodsFacade` could return the food count and the collection of foods: 

```ruby
def self.food_search(ingredient)
    result = FoodsService.get_data("search?query=#{ingredient}")
    foods = result[:foods][1..10].map do |food_data|
      Food.new(food_data)
    end
   [result[:totalHits], foods]
end
```

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,13 @@
...
+    result[:totalHits]
+  end
+
+  def self.first_ten_foods(ingredient)
+    result = FoodsService.get_data("search?query=#{ingredient}")
+    result[:foods][1..10].map do |food_data|
+      Food.new(food_data)
+    end
+  end
+end
```
megstang, line 13: Currently, you're passing the endpoing to the `get_data` method, but I think you can let your service be responsible for knowing which endpoint to call. Call a method such as `food_ingredient_search`, and let that method be responsible for calling to that specific endpoint. Your facade should just call to the service method that will make that call for you, your facade doesn't need to know the endpoint.

app/services/foods_service.rb
```diff
@@ -0,0 +1,15 @@
+class FoodsService
+
+  def self.conn
+    Faraday.new(url: 'https://api.nal.usda.gov/fdc/v1/foods/') do |faraday|
+      faraday.params['api_key'] = ENV['api_key']
+    end
+  end
+
+  def self.get_data(url)
+    response = conn.get(url)
```
megstang, line 10: you can also pass your parameter for your search ingredient in a block just like you did for the api key on line 5 above.

spec/facades/foods_facade_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+  it 'returns the count of foods that are returned from the search' do
+    expect(FoodsFacade.total_items("Turkey")).to be_an(Integer)
+    expect(FoodsFacade.total_items("Turkey")).to eq(5380)
+  end
+
+  it 'returns the first ten foods as results' do
+    expect(FoodsFacade.first_ten_foods("Turkey")).to be_an(Array)
+    expect(FoodsFacade.first_ten_foods("Turkey").length).to eq(10)
+  end
```
megstang, line 12: I would also check here that the method returns a collection of food objects -- you don't need to test the attributes of those objects, but just that they are food objects.

megstang (commented):
@acmassey3698 Good job on the mid mod. I left a few comments for you to look over, please let me know if you have any questions. I'll follow up via slack before the end of the day with some more feedback.

PR #41, opened 2021-10-25
Laura Court

Lmcourt:


app/poros/food_poro.rb
```diff
@@ -0,0 +1,12 @@
+class FoodPoro
+  attr_reader :fcdid, :description, :brand, :ingredients, :gtinupc, :total_items
+
+  def initialize(data)
+    @fcdid = data[:fcdId]
+    @description = data[:description]
+    @brand = data[:BrandOwner]
+    @ingredients = data[:ingredients]
+    @gtinucp = data[:gtinUpc]
```
epintozzi, line 9: I'm not sure if this is related to the issue you were having with your facade, but I believe there's a typo here and the c and p are reversed. Should be `@gtinupc`

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,5 @@
+class FoodsController < ApplicationController
+  def index
+    @foods = FoodFacade.foods_by_keyword(params[:search])
```
epintozzi, line 3: We should be using the param `:q` here. This is defined in our `welcome.html.erb` on line 9: `<%= text_field_tag(:q) %>`

You could also verify by dropping a pry in your controller and checking the params to see how your search term is coming in if you aren't sure how to find the param key in the view.

Revised, this should be: `@foods_list = FoodFacade.foods_by_keyword(params[:q])`

epintozzi:
Nice work. A couple little things I think would've had this working for you. I'll follow up with more feedback via slack shortly. Let me know if you have questions.

PR #42, opened 2021-10-25
Michael Mattson

michaelpmattson:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,6 @@
+class FoodsController < ApplicationController
+  def index
+    @total_foods = FoodsFacade.total_foods(params[:q])
+    @foods       = FoodsFacade.first_ten_foods(params[:q])
```
megstang, line 4: You mentioned in your submission that you'd like to refactor so that you're only calling the api endpoint once. 

You could do something like this: 
`@total_foods, @foods = FoodsFacade.foods_for_ingredient(params[:q])`

and then in a `foods_for_ingredient` method in your Facade, do something like: 

```ruby 
 data = FoodsService.foods_query(query)

    foods_data = data[:foods]
    foods = foods_data[0..9].map do |food_data|
      Food.new(food_data)
    end
    [foods_data[:totalHits], foods]
```

this will respectfully assign the variables in the controller.

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,16 @@
...
+    data = FoodsService.foods_query(query)
+
+    data[:totalHits]
+  end
+
+  def self.first_ten_foods(query)
+    data = FoodsService.foods_query(query)
+
+    foods_data = data[:foods]
+    foods_data[0..9].map do |food_data|
```
megstang, line 12: The api has a perPage parameter that it takes, so you could set that perPage parameter to 10 to be sure you only get 10 results back with one call.

app/poros/food.rb
```diff
@@ -0,0 +1,15 @@
...
+              :ingredients
+
+  def initialize(food_data)
+    @fdc_id      = food_data[:fdcId]
+    @gtin_upc    = food_data[:gtinUpc]
+    @description = food_data[:description]
+    @brand_owner = food_data[:brandOwner]
+    @ingredients = food_data[:ingredients]
+  end
+end
```
megstang, line 15: 👍

app/services/foods_service.rb
```diff
@@ -0,0 +1,19 @@
...
+  def self.conn
+    Faraday.new(url: "https://api.nal.usda.gov/fdc") do |faraday|
+      faraday.headers["X-Api-Key"] = ENV['foods_key']
+    end
+  end
+
+  def self.parse_json(response)
+    JSON.parse(response.body, symbolize_names: true)
+  end
+end
```
megstang, line 19: 👍

megstang (commented):
@michaelpmattson Really nice job. I left a few comments for you to look over. Let me know if you have any questions. I know you know you're not at 100% testing yet, so you may already have this thought, but make sure you're also testing facades ;) I'll follow up via a slack dm before the end of the day with more feedback.

PR #43, opened 2021-10-25
Khoi Nguyen

khoifishpond:
**kill me**


config/routes.rb
```diff
@@ -1,3 +1,4 @@
 Rails.application.routes.draw do
   root "welcome#index"
+  resources :foods
```
epintozzi, line 3: Since we are only using index, we could write this as `resources :foods, only: :index`

app/services/food_service.rb
```diff
@@ -0,0 +1,7 @@
+class FoodService < ApiService
+  BASE_URL = 'https://api.nal.usda.gov/fdc'
+
+  def self.search_food(ingredient)
+    get_data("#{BASE_URL}/v1/search?#{ingredient}&api_key=#{ENV['food_api']}")
```
epintozzi, line 5: We also need to include the `query` key when passing our search term. So this should read more like:
`/search?query=#{ingredient}&api_key=#{ENV['food_api']}`

app/facades/food_facade.rb
```diff
@@ -0,0 +1,8 @@
+class FoodFacade
+  def self.search(food)
+    json = FoodService.search_food(food)
+    json.map do |result|
+      Food.new(result)
+    end
```
epintozzi, line 6: I think you will have some issues with this mapping because we need to first pull the list of foods out of the result. If you change line 3 to:
`json = FoodService.search_food(food)[:foods]` that should help get you on your way.

PR #44, opened 2021-10-25
Luis

larroyo1:


app/services/api_service.rb
```diff
@@ -0,0 +1,13 @@
+class APIService
+  # URL = 'https://fdc.nal.usda.gov/api-spec/fdc_api.html#/FDC/getFoodsSearch'
+
+  def self.parse_json(response)
+    JSON.parse(response,body, symbolize_names: true)
+  end
+
+  def self.conn
+    Faraday.new('https://fdc.nal.usda.gov/api-spec/fdc_api.html#/FDC') do |f|
```
megstang, line 9: This url that you have here is for the documentation for the api. You'll need to look at the documentation to see what the api's domain is.

megstang: Looks like there's a drop down that lists the servers you could connect to. The first one is the url you should be using to make your connection, and then you'll need to look down the list on the docs to identify the best endpoint to use.

app/services/food_service.rb
```diff
@@ -0,0 +1,9 @@
+class FoodService < APIService
+
+  def self.search_foods_by_name(name)
+    response = conn.get('/getFoodsSearch') do |f|
```
megstang, line 4: `/getFoodsSearch` is not an endpoint listed in the docs. Look at the list (green and blue rows) here to see the endpoints you could get data from. https://fdc.nal.usda.gov/api-spec/fdc_api.html#/FDC

megstang (commented):
@dat1guyluigi Hey Luis, I left a few comments for you to look over. I'll follow up with more feedback via slack before the end of the day.

PR #45, opened 2021-10-25
Sami P

sami-p:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,7 @@
+class FoodsController < ApplicationController
+
+  def index
```
epintozzi, line 3: `get_10` and `total_hits` the same method in the service (`search_ingredients`) so we don't really need to have them separated in the Facade and called twice. You can handle the search hits data from the call you make for search and assign them to appropriate variables after the single call.

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+RSpec.describe 'Food Index Page' do
+  it 'returns first 10 foods and attributes' do
+    visit root_path
+
+    fill_in :q, with: 'sweet potatoes'
+    click_on 'Search'
+
+    expect(current_path).to eq(foods_path)
+    expect(current_page).to have_content('BAKO SWEET')
```
epintozzi, line 11: This spec fails for me but works if I use `page` instead of `current_page`

epintozzi:
Great job! I'll follow up in slack with more feedback shortly. Let me know if you have questions.

PR #46, opened 2021-10-25
Erin Quinn

equinn125:


app/facades/food_facade.rb
```diff
@@ -0,0 +1,8 @@
+ class FoodFacade
+   def self.ten_food_search(search_param)
+    foods_data =  FoodService.food_search(search_param)[:foods]
+     foods_data[0..9].map do |food_data|
```
epintozzi, line 4: I think this should work without the `[0..9]` because you are passing the param for `pageSize` with a limit of 10 when you make the API call in your `FoodService`. There's no harm in doing it here, however it isn't necessary if the API is honoring your pageSize limit of 10. If you prefer to do it here, you could also drop the `pageSize` param but I it's probably more efficient to limit it at the Service level.

app/services/food_service.rb
```diff
@@ -0,0 +1,20 @@
+class FoodService
+
+  def self.food_search(food)
+    response = conn.get('/fdc/v1/foods/search?') do |faraday|
+    faraday.params[:query] = "#{food}"
+    faraday.params[:pageSize] = 10
```
epintozzi, line 6: I like that you are limiting to 10 here. It's more efficient than retrieving everything and paring it down later if we know we do not need more data.

config/routes.rb
```diff
@@ -1,3 +1,5 @@
 Rails.application.routes.draw do
   root "welcome#index"
+
+  resources :foods
```
epintozzi, line 4: We should limit to only the routes we need, so I'd write this instead as:
`resources :foods, only: :index`

spec/facades/food_facade_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'rails_helper'
+
+describe FoodFacade do
+  it 'returns colleciton of ten Food objects' do
+    foods = FoodFacade.ten_food_search("sweet potatoes")
+    expect(foods).to be_a Array
+    expect(foods.first).to be_a Food
+    expect(foods.count).to eq(10)
+  end
```
epintozzi, line 9: Nice job with this test. I like that you are testing that we are returning a `Food` object rather than the specific attributes of the food since the Facade doesn't care about that.

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'rails_helper'
+
+RSpec.describe 'food index' do
+  it 'allows the user to search for foods' do
+    visit root_path
+    fill_in :q, with: "sweet potatoes"
+    click_button 'Search'
+    expect(page.status_code).to eq(200)
+  end
```
epintozzi, line 9: This test could be a little more thorough. Ideally it would've looked a lot like the user story, at the very least verifying that our search term is on the page and that our current path is `/foods`

epintozzi:
Nice job on your midmod. I will follow up via slack by the end of the day with more feedback. Let me know if you have any questions!

PR #47, opened 2021-10-25
Dane Brophy 🎃

renamabo:
Felt I did well. Ran out of time, else I believe I'd be able to figure out why the controller action isn't allowing access to all params.

### DIFF
--- a/app/controllers/foods_controller.rb
+++ b/app/controllers/foods_controller.rb
@@ -1,6 +1,7 @@
 class FoodsController < ApplicationController
 
   def index
-    require "pry"; binding.pry
+    @foods = FoodsFacade.foods_search(params[:search_params])
+    @food_count = FoodsService.total_foods("sweet%20potatoes")[:totalHits]
   end
 end
--- a/app/controllers/foods_controller.rb
+++ b/app/controllers/foods_controller.rb
@@ -1,4 +1,6 @@
 class FoodsController < ApplicationController
 
-  def index; end
+  def index
+    @food = FoodsFacade.foods_search(food_params)
+  end
 end
--- a/app/controllers/foods_controller.rb
+++ b/app/controllers/foods_controller.rb
@@ -1,0 +1,4 @@
+class FoodsController < ApplicationController
+
+  def index; end
+end
--- a/app/controllers/foods_controller.rb
+++ b/app/controllers/foods_controller.rb
@@ -1,6 +1,6 @@
 class FoodsController < ApplicationController
 
   def index
-    @food = FoodsFacade.foods_search(food_params)
+    require "pry"; binding.pry
   end
 end
--- a/app/services/foods_service.rb
+++ b/app/services/foods_service.rb
@@ -1,10 +1,16 @@
 class FoodsService
   class << self
-    def search_foods(query)
-      response = conn.get("/fdc/v1/foods/search?query=#{query}&pageSize=25&sortOrder=asc")
+
+    def total_foods(query)
+      response = conn.get("/fdc/v1/foods/search?query=#{query}&pageSize=10&sortOrder=asc")
       parse_json(response)
     end
 
+    def search_foods(query)
+      response = conn.get("/fdc/v1/foods/search?query=#{query}&pageSize=10&sortOrder=asc")
+      body = parse_json(response)
+      body[:foods]
+    end
 
     private
 
--- a/app/services/foods_service.rb
+++ b/app/services/foods_service.rb
@@ -1,0 +1,21 @@
+class FoodsService
+  class << self
+    def search_foods(query)
+      response = conn.get("/fdc/v1/foods/search?query=#{query}&pageSize=25&sortOrder=asc")
+      parse_json(response)
+    end
+
+
+    private
+
+    def conn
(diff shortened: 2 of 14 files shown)


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,7 @@
+class FoodsController < ApplicationController
+
+  def index
+    @foods = FoodsFacade.foods_search(params[:search_params])
```
megstang, line 4: The search parameter coming in is :q, not :search_params. If you take a look at the form that's on the welcome page, you can see the naming of the form field. This would make sense why your index action isn't able to read in a value from the parameters.

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,7 @@
...
+    @food_count = FoodsService.total_foods("sweet%20potatoes")[:totalHits]
```
megstang, line 5: both `FoodsService.total_foods(..)` and `FoodsFacade.foods_search(..)` call to the same API endpoint in the service, so they're both getting back food AND totalHits information. You could just have one method return both of those pieces of data so that you don't need to call to the same endpoint twice. Here, in your index action, you could do something like: 

```ruby
@foods, @food_count = FoodsFacade.foods_search(params[:q])
```

and then in your `foods_search` method, you could return both values in an array, which will respectively assign the variables above. 

```ruby
def foods_search(query)
      data = FoodsService.search_foods(query)
      foods = data[:foods].map do |food_data|
        Food.new(food_data)
      end
     [foods, data[:totalHits]]
    end
```

app/services/foods_service.rb
```diff
@@ -0,0 +1,27 @@
+class FoodsService
+  class << self
+
+    def total_foods(query)
+      response = conn.get("/fdc/v1/foods/search?query=#{query}&pageSize=10&sortOrder=asc")
```
megstang, line 5: Nice use of the pageSize parameter on this endpoint 👍

app/services/foods_service.rb
```diff
@@ -0,0 +1,27 @@
...
+      parse_json(response)
+    end
+
+    def search_foods(query)
+      response = conn.get("/fdc/v1/foods/search?query=#{query}&pageSize=10&sortOrder=asc")
+      body = parse_json(response)
+      body[:foods]
```
megstang, line 12: I'd save this line for the facade since your service methods should just be making calls to the api -- worry about grabbing the data you want from within your facade.

app/services/foods_service.rb
megstang: You can also use block parameters for the query, pageSize, and sortOrder, just like you did in the conn method for the api_key parameter.

app/views/welcome/index.html.erb
```diff
@@ -6,7 +6,7 @@
     <ul class="nav navbar-nav">
       <%= form_tag :foods, method: :get, class: "form-inline" do %>
         <div class="form-group search-field">
-          <%= text_field_tag(:q) %>
+          <%= text_field_tag(:search_params) %>
```
megstang, line 9: 👍 Forget about my comment above 😉

spec/facades/foods_facade_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+  it '#foods_search' do
+
+    foods = FoodsFacade.foods_search("sweet%20potatoes")
+
+    expect(foods).to be_an(Array)
+    expect(foods.first).to be_a(Food)
+    expect(foods.first.gtin_code).to be_a(String)
+    expect(foods.first.description).to be_a(String)
+    expect(foods.first.brand_owner).to be_a(String)
+    expect(foods.first.ingredients).to be_a(String)
```
megstang, line 13: If you test your food poro, there's no need for these four lines here.

megstang (commented):
@danembb Super close to finishing up all the functionality! I also think that if you had just a few more minutes you could have wrapped this up. I put a few comments on your code. Please take a look and let me know if you have any questions. I'll follow up later in the day via slack with more feedback.

spec/poros/food_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'rails_helper'
+
+RSpec.describe 'food poro' do
+  it 'exists and has attributes' do
+    oat_milk = Food.new(gtin_code: "8239084093", description: "I cant believe its not milk!", brand_owner: "the oat people", ingredients: "oats!")
```
megstang, line 5: Your food poro takes in data that comes back from the api, so it's a hash of data that looks something like this: 

```
{:fdcId=>1929370,
  :description=>"SWEET POTATOES",
  :lowercaseDescription=>"sweet potatoes",
  :dataType=>"Branded",
  :gtinUpc=>"076700002019",
  :publishedDate=>"2021-07-29"
......
```

You're trying to make a food object by setting the attributes directly. In your food poro, you're grabbing the keys of `brandOwner` and `gtinUpc` in the initialize method, and those are not the keys that you're passing in in your test, but those ARE the keys that will be coming through when you're passing the api response along to the initialize method to create your Food objects in your facade. SO, change up your Food.new(...) argument to be the hash that you'd be passing in from the facade, and this should be all set to go!

PR #48, opened 2021-10-25
Isika Powers

Isikapowers:


config/routes.rb
```diff
@@ -1,3 +1,5 @@
 Rails.application.routes.draw do
   root "welcome#index"
+
+  resources :foods
```
epintozzi, line 4: We can limit this to just the index since that's the only route we are using
`resources :foods, only: :index`

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,12 @@
+<h1>List of Foods:</h1>
+
+<% @foods.take(10).each do |food| %>
```
epintozzi, line 3: This works for showing only ten foods. We can also use the `pageSize` param when making our API call to return only 10 foods. This is more efficient because we aren't creating a bunch of extra poros, will speed up our request/response, and also puts less load on the service providing our data.

app/facades/food_facade.rb
```diff
@@ -0,0 +1,10 @@
+class FoodFacade
+  def self.ten_foods(item)
```
epintozzi, line 2: Since we're actually returning all foods here (not just ten), I'd probably name this method something else for clarity.

spec/poros/food_info_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+      potatos = VCR.use_cassette("food_info") do
+        FoodFacade.ten_foods("sweet potatos")
+      end
+
+      expect(potatos[0]).to be_a FoodInfo
+      # expect(potatos[0].upc_code).to eq()
+      expect(potatos[0].description).to eq("Bread, sweet potato")
+      expect(potatos[0].brand_owner).to eq()
+      expect(potatos[0].ingredients).to eq()
```
epintozzi, line 16: These could be a little more robust to test for attribute values

epintozzi:
Nice job! I'll follow up via slack later today with more feedback. Let me know if you have any questions!

PR #50, opened 2021-10-25
Rowan

RowanDW:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,8 @@
+class FoodsController < ApplicationController
+  def index
+    search = FoodFacade.search(params[:q])
+    @hits = search[:total_hits]
+    @foods = search[:foods]
+  end
```
megstang, line 6: Nice! I really like how you did this here. Since you have the `search` method returning a hash with both values in it, you might consider just passing that object to the view to stick with the ruby convention of having one object passed to the view. Then, in your view, you could access the data you needed by calling `@search[:total_hits]` or `@search[:foods]`, although you may change to rename @ search to something else.

app/facades/food_facade.rb
```diff
@@ -0,0 +1,14 @@
+class FoodFacade
+  def self.search(query)
+    query.gsub!(' ', '+')
+    response = FoodService.search(query)
+require "pry"; binding.pry
+    foods = response[:foods].slice(0, 10)
```
megstang, line 6: You could utilize the perPage param with the api to only get back 10 responses for that first page. That way, you wouldn't haave to limit it here.

app/poros/food.rb
```diff
@@ -0,0 +1,13 @@
...
+              :brandOwner,
+              :ingredients
+
+  def initialize(data)
+    @gtinUpc          = data[:foodCode]
+    @description      = data[:description]
+    @brandOwner       = data[:brand_owner]
+    @ingredients      = data[:ingredients]
+  end
+end
```
megstang, line 13: The reason why you aren't seeing the brand owner and the foodCode get displayed is because those are not how the keys are returned from the service. I believe the keys are :gtinUpc and :brandOwner.

app/services/food_service.rb
```diff
@@ -0,0 +1,17 @@
+class FoodService
+
+  def self.search(query)
+    response = conn.get("/fdc/v1/foods/search?query=#{query}")
```
megstang, line 4: So on here, if you also added a `&perPage=10`, then you'll only get back 10 food results, but you'll still have access to the totalHits value.

megstang (commented):
@RowanDW Really great job on this. I'll follow up via slack with some more feedback before the end of the day. I think the reason why it wasn't fully functional was because you were accessing the wrong (misspelled) keys in the Food Poro. Once I fixed that, your mid mod was fully functional.

PR #51, opened 2021-10-25
JJ

JeJones21:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,9 @@
+class FoodsController < ApplicationController
+
+  def index
+   @results = FoodFacade.food_search(params[:q])
+   @result_count = results[:result_count]
+   @foods = results[:top_ten]
```
epintozzi, line 6: We need to use `@results` here on lines 5 and 6 (so `@results[:result_count]` and `@results[:top_ten]`). This might resolve the error you were running into with your facade?

app/facades/food_facade.rb
```diff
@@ -0,0 +1,11 @@
+class FoodFacade
+
+  def self.food_search(ingredient)
+   results = FoodService.search_by_ingredient(ingredient)
+   {
+     result_count: results[:totalHits],
+     top_ten: results[:foods].map { |result| Food.new(result) }
+   }
+  end
```
epintozzi, line 9: I really like this method. You are only making one call to your api (which is good and preferred) and you are elegantly assigning the things that need to be assigned for you to use later.

poros/food.rb
```diff
@@ -0,0 +1,14 @@
+class Food
```
epintozzi, line 1: Poros should be within the app directory (`app/poros/food.rb`)

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,9 @@
+<h1>Get In My Belly</h1>
+
+<h2>Search Results: <%= @result_count %></h2>
+<% @foods.each do |food| %>
+  <h4><%= food.description %></h4><br>
+  <p>Brand: <%= food.brandOwner %></p>
+  <p>Ingredients: <%= food.ingredients %></p>
+  <p>UPC Code: <%= food.gtinUpc%></p>
```
epintozzi, line 8: Because this attribute is all lowercase in your poro, I believe it needs to be lowercase here (`gtinupc` vs `gtinUpc`)

poros/food.rb
epintozzi: And if the above comment in the controller didn't resolve the issue you were having with the Facade, moving this to `app` might do it. It just depends what the error was that you were getting.

epintozzi:
Nice job! I'll follow up with more feedback via slack before the end of the day. Let me know if you have any questions.

app/controllers/foods_controller.rb
JeJones21: Still getting NameError uninitialized constant. When I pry I can see my response as requested. But I keep getting this error
NameError: uninitialized constant FoodFacade::Food.. something between my facade and poro is disconnected and I have not been able to resolve.

poros/food.rb
JeJones21: OH my Erin that was it! Fixed the entire issue... Thank you

PR #52, opened 2021-10-25
Ozzie Osmonson

ozzman84:
Wasted too much time on small testing mistakes and it ate up most of my time.

### DIFF
--- a/app/services/food_service.rb
+++ b/app/services/food_service.rb
@@ -1,0 +1,7 @@
+class FoodsService
+  class << self
+    def foods_by_name(search_string)
+      
+    end
+  end
+end
--- a/spec/foods/index_spec.rb
+++ b/spec/foods/index_spec.rb
@@ -1,0 +1,18 @@
+require 'rails_helper'
+
+RSpec.describe 'Food Index' do
+  describe 'As a User' do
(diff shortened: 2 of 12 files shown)


app/services/food_service.rb
```diff
@@ -0,0 +1,7 @@
+class FoodsService
```
epintozzi, line 1: Your filename here should be `/foods_service.rb` plural because your class name is plural. I'm not sure if this is one of the errors you were running into, but it would've probably become one at some point. It would've read as an uninitialized constant error.

spec/foods/index_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'rails_helper'
```
epintozzi, line 1: For this spec, I believe if you'd have put `/foods` in a `features` directory (like `/spec/features/foods/index_spec.rb`), this would've resolved your error about an undefined path. Rails is very particular about how it expects to find things and can result in some very frustrating errors when we don't do think how Rails likes it.

epintozzi:
Thanks for you submission. I am sure these were very frustrating errors to run in to. I'll follow up with some additional feedback via slack.

PR #53, opened 2021-10-25
Christina Delpone

cdelpone:


app/facades/foods_facade.rb
```diff
@@ -0,0 +1,9 @@
+class FoodsFacade
+  def self.search_foods_by_ingredient(ingredient)
+    food_data = FoodsService.get_data("search?query=#{ingredient}")
+
+    food_data.map do |data|
+      Food.new(data)
+    end
+  end
```
megstang: Currently, you're passing the endpoint to the `get_data` method, but I think you can let your service be responsible for knowing which endpoint to call. Call a method `such as food_ingredient_search`, and let that method be responsible for calling to that specific endpoint. Your facade should just call to the service method that will make that call for you, your facade doesn't need to know the endpoint.

megstang: The response that you get from that endpoint will also return a totalHits value that you can grab and pass back to the controller action to then get sent to the view. You could, grab that value and return it along with the `food_data` collection: 

```ruby
  def self.search_foods_by_ingredient(ingredient)
    food_data = FoodsService.get_data("search?query=#{ingredient}")

    food_objects = food_data.map do |data|
      Food.new(data)
    end
   {total_hits: food_data[:totalHits], foods: food_objects}
  end
```

app/poros/food.rb
```diff
@@ -0,0 +1,8 @@
+class Food
+  attr_reader :fdcid
+
+  def initialize(data)
+    @fdcid = data[:fdcid]
+
+  end
+end
```
megstang: I think you were headed here, but your `data` argument should also have the other attributes that you want to be able to read from the view. You can pull out and assign those values here.

app/services/foods_service.rb
```diff
@@ -0,0 +1,14 @@
+class FoodsService
+  def self.conn
+    Faraday.new(url: 'https://api.nal.usda.gov/fdc/v1/foods/') do |faraday|
+      faraday.params['api_key'] = ENV['food_key']
+    end
+  end
+
+  def self.get_data(url)
+    response = conn.get(url)
```
megstang, line 9: currently, your url contains the parameters for the ingredient search. If you structured this method such that it knew the endpoint to call to (see the first comment on the FoodFacade for more reference on this suggestion), you could pass the ingredient into teh method, and use a block to set the parameter, just like how you set the api key on line 4.

spec/services/food_service_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    expect(response[0]).to be_a(Hash)
+    expect(response).to be_an(Array)
+    expect(response[0][:fdcId]).to be_an(Integer)
+    expect(response[0][:description]).to be_a(String)
+    expect(response[0][:dataType]).to be_a(String)
+    expect(response[0][:publicationDate]).to be_a(String)
+    expect(response[0][:foodCode]).to be_a(String)
+    expect(response[0][:foodNutrients]).to be_an(Array)
+    expect(response[0][:foodNutrients][0]).to be_a(Hash)
+  end
```
megstang: This endpoint that you're passing to the get_data method is not the endpoint that you are callign to in your facade.

megstang: Instead, I would pass the endpoint you're using in your facade and testing that response instead. 

`FoodsService.get_data("search?query=sweet+potato")`

megstang (commented):
@cdelpone I left a few comments for you to look over. Please let me know if you have any questions. I will follow up with more feedback via slack before the end of the day.

app/controllers/welcome_controller.rb
```diff
@@ -1,4 +1,5 @@
 class WelcomeController < ApplicationController
   def index
+    @foods = FoodsFacade.search_foods_by_ingredient(params[:search])
```
megstang, line 3: I believe the search parameter that's coming in is :q, not :search

PR #54, opened 2021-10-25
Kevin Mugele Mid Mod

KevinMugele:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,6 @@
+class FoodsController < ApplicationController
+  def index
+    @foods = FoodFacade.search_results(params[:q])
+    @total_hits = FoodFacade.total_results(params[:q])
+  end
```
megstang, line 5: Both of these facade methods, `search_results` and `total_results` are eventually calling to the same service method, which means it's calling to the same endpoint. Since both totalHits AND food data is returned by calling to that endpoint once, let's consider an approach that would pull out the data you need for your view with just ONE api call: 

```ruby
@foods, @total_hits = FoodFacade.search_results(params[:q])
```

So, in your `search_results` action in your facade, you could do something like: 

```ruby
def search_results(search)
      data = FoodService.food_search_results(search)
      foods = data[:foods].map do |food_data|
        Food.new(food_data)
      end
      [foods, data[:totalHits]]
    end
```

app/facades/food_facade.rb
```diff
@@ -0,0 +1,24 @@
...
+        #currently listing results where the searched item is listed in description or Ingredients
+
+      # if !food[:ingredients].nil?
+      #   if food[:ingredients].include?(search)
+      #     Food.new(food)
+      #   end
+      # end
+
+    end
+    foods.take(10)
```
megstang, line 17: Instead of iterating and creating food objects out of EVERYTHING, and THEN taking the first 10, you could either just iterate through the first 10 in the collection OR, you could use the perPage parameter that the api docs mention to just get 10 foods back.

app/poros/food.rb
```diff
@@ -0,0 +1,11 @@
...
+  attr_reader :description, :gtin, :brand_owner, :ingredients
+
+  def initialize(attributes)
+  
+    @description = attributes[:description]
+    @gtin = attributes[:gtinUpc]
+    @brand_owner = attributes[:brandOwner]
+    @ingredients = attributes[:ingredients]
+  end
+end
```
megstang, line 11: 👍

app/services/foods_service.rb
```diff
@@ -0,0 +1,14 @@
+class FoodsService
+  def self.food_search_results(search)
+      response = conn.get("search?query=#{search}&api_key=#{ENV['food_api_key']}")
```
megstang, line 3: A more conventional way of setting parameters using Faraday as your client, is to assign them within a block: 

```ruby 
conn.get("search") do |req|
   req.params[:query] = search 
   req.params[:api_key] = ENV['food_api_key']
end 
```

app/services/foods_service.rb
```diff
@@ -0,0 +1,14 @@
...
+      parse_json(response)
+  end
+
+  def self.conn
+      Faraday.new(url: "https://api.nal.usda.gov/fdc/v1/foods/")
```
megstang, line 8: I would actually move `/fdc/v1/foods/` into the get request on line 3 since that's part of the endpoint that you're hitting for this server.

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,18 @@
...
+  </tr>
+<% @foods.each do |food| %>
+  <tr>
+    <td><%= food.description %></td>
+    <td><%= food.gtin %></td>
+    <td><%= food.brand_owner %></td>
+    <td><%= food.ingredients %></td>
+  </tr>
+  <% end %>
+</table><br>
```
megstang, line 18: 👍

spec/features/food/index_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      click_button "Search"
+      results = FoodFacade.total_results('sweet potato')
+      foods = FoodFacade.search_results('sweet potato')
+      food = foods.first
+
+      expect(page.status_code).to eq 200
+      expect(page).to have_content(results)
+      expect(foods.size).to eq(10)
+      expect(page).to have_content(food.description)
+    end
```
megstang, line 35: Test to make sure the attributes you're expecting to be present are appearing on the page -- so that'll include a few more than just the description.

megstang (commented):
@KevinMugele Really great job! I left a few comments for you to look over. Please let me know if you have any questions. i'll follow up via slack DM before the end of the day with some more feedback.

PR #58, opened 2022-02-23
Darren Kulback

dkulback:


app/clients/foods_client.rb
```diff
@@ -0,0 +1,15 @@
+class FoodsClient
```
epintozzi, line 1: I like the setup of this class as well as the guard clause on line 5

app/services/foods_servicer.rb
```diff
@@ -0,0 +1,18 @@
...
+  def search(food)
+    results = client.search_foods(food)[:foods].map do |data|
+      Food.new(data)
+    end
+    results[0..9]
+  end
+
+  def get_total_hits(food)
+    client.search_foods(food)[:totalHits]
+  end
```
epintozzi, line 17: Ultimately these methods are hitting the same endpoint twice. Instead you could refactor this to something like:
```ruby
  def search(food)
    results = client.search_foods(food)
    foods = results[:foods].map do |data|
      Food.new(data)
    end
    [results[:totalHits], foods[0..9]]
  end
  ```
  
  And then in your controller:
  `    @results, @foods = FoodsServicer.new.search(params[:q])`
  
  This should allow you to get rid of the `get_total_hits` method altogether.

epintozzi (commented):
Great work and really nice testing!

PR #59, opened 2022-02-23
David Kassin

dkassin:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,10 @@
+class FoodsController < ApplicationController
+  def index
+    @food = params[:q]
+    if @food.empty?
+      redirect_to root_path
+    else
+      @foods = FoodFacade.all_foods(@food)
+    end
+  end
```
epintozzi, line 9: This is awesome. I really love seeing that you've accounted for this edgecase here.

app/facades/food_facade.rb
```diff
@@ -0,0 +1,10 @@
+class FoodFacade
+  def self.all_foods(food)
+    json_data = FoodService.foods(food)
+    food_hits = json_data[:totalHits]
```
epintozzi, line 4: Nice work using the one API all to also get the count instead of making the API call a second time to accomplish this.

app/poros/food.rb
```diff
@@ -0,0 +1,11 @@
+class Food
+  attr_reader :hits, :gtinUpc, :description, :brandOwner, :ingredients
+
+  def initialize(data, hits)
+    @hits = hits
```
epintozzi, line 5: I don't mind that you're storing the count on every `Food` item, however an argument against this is that is repetitive to store the same piece of unchanging data each time and takes up space in memory. It would be something more concerning in a larger app or if we were handling very large amounts of data.

spec/features/search_foods_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      expect(page.status_code).to eq 200
+
+      expect(page).to have_content("Total Number of items returned by search: 46770")
+      expect(page).to have_content("GTIN/UPC: 819614010394")
+      expect(page).to have_content("Brand Owner: Country Sweet Produce")
+      expect(page).to have_content("Ingredients: SWEET POTATOES")
+    end
+
+    it 'has sad path for empty search' do
```
epintozzi, line 20: 🎉

app/facades/food_facade.rb
```diff
@@ -0,0 +1,10 @@
...
+
+    foods = json_data[:foods].map do |food|
+      Food.new(food, food_hits)
+    end
```
epintozzi, line 8: In relation to the comment below on your poro, if instead of saving the total hits to each `Food` instance, you could return the hits and foods in an array at the end of this method, like this:

```ruby
  def self.all_foods(food)
    json_data = FoodService.foods(food)
    food_hits = json_data[:totalHits]

    foods = json_data[:foods].map do |food|
      Food.new(food, food_hits)
    end
    [food_hits, foods]
  end
end
```

And then in your controller do something like this:
```ruby
...code
  else
     @food_hits, @foods = FoodFacade.all_foods(@food)
  end
```

epintozzi (commented):
Great work! This is well written and your testing is super thorough 🎉

PR #60, opened 2022-02-23
John Napier Mid Mod

JCNapier:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,8 @@
+class FoodsController < ApplicationController 
+  def index 
+    if params["q"].present?
+      @foods = UsdaFacade.foods_by_keyword(params["q"])
+      @total_hits = UsdaService.foods_by_keyword(params["q"])[:totalHits]
```
jamisonordway, line 5: Instead of calling the same method twice (and thus hitting the endpoint twice), line 5 could become
`@total_hits = @foods[:totalHits]`

app/services/usda_service.rb
```diff
@@ -0,0 +1,13 @@
+class UsdaService 
+  def self.conn 
+    Faraday.new(url: 'https://api.nal.usda.gov') do |f|
+      f.params['api_key'] = ENV['usda_key']
+    end
+  end
+
+  def self.foods_by_keyword(food)
+    response = conn.get("/fdc/v1/foods/search?query=#{food}&dataType=&pageSize=10&pageNumber=1")
```
jamisonordway, line 9: This request is imposing a page size limit of 10, which will also affect the @total_hits ivar that you're passing to the view.

Instead of limiting the results to 10 here, this service could get all the foods that match the query and the facade could be responsible for creating POROs out of the first 10. That might look something like this:

```
# in UsdaFacade

def self.foods_by_keyword(food, result_limit = 10)
  json = UsdaService.foods_by_keyword(food)
  food_results = json[:foods].map(&:food).take(result_limit)
  foods = food_results.map do |food|
    Food.new(food)
  end
  
  {total_hits: json[:totalHits], foods: foods}
end
```

jamisonordway (commented):
Good stuff! I appreciate that you tested everything- the index page, PORO, service, and facade.
I did my best to answer your question about total number of hits in my comments, let me know if you'd like to discuss further.

PR #62, opened 2022-02-23
Dylan Harper

dylan-harper:


dylan-harper:
Forgot to leave my review!

Thought this went pretty well, I was sitting pretty comfortable around the 90min mark, until I realized I was supposed to filter the items by ingredients. Threw that in last minute and it seemed to work. If it was a db search, I would have been more comfortable using AR or SQL to filter them but figured simple ruby would do. Would have liked to expand more on my service/facade/poro specs, but glad I got (what I believe to be) a finished user story.

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,14 @@
+class FoodsController < ApplicationController
+
+  def index
+    @results = FoodSearchFacade.search_foods(food_params[:q])
```
epintozzi, line 4: Since we aren't trying to write to our db, we don't need to use safe params here. You should be able to pass `params[:q]` directly.

app/facades/food_search_facade.rb
```diff
@@ -0,0 +1,22 @@
...
+    results = service.get_foods(query)
+    poros = results[:foods].map { |data| Food.new(data) }
+
+    foods = []
+    poros.each do |food|
+      if food.ingredients != nil
+        if food.ingredients.include? query.upcase
+          foods << food
+        end
+      end
```
epintozzi, line 13: This is an interesting interpretation of the user story! It totally works and it also would've been fine if you opted to just return the results as is even if there would be some gaps in the data.

app/services/food_search_service.rb
```diff
@@ -0,0 +1,9 @@
+class FoodSearchService
+
+  def get_foods(query)
+    food_name = ERB::Util.url_encode query
+    response = Faraday.get("https://api.nal.usda.gov/fdc/v1/foods/search?query=#{food_name}&api_key=#{ENV['foods_api_key']}")
+    parsed = JSON.parse(response.body, symbolize_names: true)
```
epintozzi, line 6: Tiny thing, but it doesn't look like you are using the variable `parsed` anywhere so this line could just be: `JSON.parse(response.body, symbolize_names: true)`

spec/facades/foods_search_facade_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    query = 'sweet potatoes'
+
+    results = FoodSearchFacade.search_foods(query)
+
+    results.each do |i|
+      expect(i).to be_a Food
+      # expect(i.brand_owner).to be_a String
+      # expect(i.description).to be_a String
+      # expect(i.gtinUpc).to be_a String
+      # expect(i.ingredients).to be_a String
```
epintozzi, line 15: I like that you moved this to the `food_spec`. The facade doesn't care to much about the actual data stored on `Food`.

epintozzi (commented):
Nice work! Your testing is also very solid and it's awesome to see that you were able to get vcr implemented for this challenge as well.

PR #63, opened 2022-02-23
Wade Naughton

WadeNaughton:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,12 @@
+class FoodsController < ApplicationController
+
+  def index
+    # food = params[:q]
+    # conn = Faraday.new(url: "https://api.nal.usda.gov")
+    # response = conn.get("/fdc/v1/foods/search?query=#{food}&api_key=#{ENV['foods_api_key']}")
+    # @json = JSON.parse(response.body, symbolize_names: true)
+    # @foods = @json[:foods]
```
epintozzi, line 8: It's a good idea to clean up old code when you're done with it.

app/services/food_service.rb
```diff
@@ -0,0 +1,12 @@
+class FoodService
+  def self.conn
+      Faraday.new("https://api.nal.usda.gov") do |faraday|
+        faraday.params['api_key'] = ENV['foods_api_key']
+      end
+    end
+
+    def self.food_info(food)
+      response = conn.get("/fdc/v1/foods/search?query=#{food}&api_key=#{ENV['foods_api_key']}")
```
epintozzi, line 9: Since you are including your api key as a param in the `conn` method on line 4, you don't need to include it here, too.

app/poros/food.rb
```diff
@@ -0,0 +1,15 @@
...
+              :brand,
+              :ingredients,
+              :gtinUpc,
+              :total
+  def initialize (foods, total)
+    @description = foods[:description]
+    @brand = foods[:brandOwner]
+    @ingredients = foods[:ingredients]
+    @gtinUpc = foods[:gtinUpc]
+    @total = total
```
epintozzi, line 13: I don't mind that you are storing the total number of foods on each `Food` object, but one argument against this is that it's unnecessary to carry that data on each `Food` and it will take up space in memory. This is a very small scope app so wouldn't matter much but on a large scale app, we'd want to consider if that's the best way to handle it.

epintozzi (commented):
Really nice work! Your api consumption with the service/facade pattern looks great and your testing is super thorough 🎉

PR #64, opened 2022-02-23
Jackson Valdez

jacksonvaldez:


app/facades/food_facade.rb
```diff
@@ -0,0 +1,15 @@
+class FoodFacade
+
+  def self.search_foods(query, page_num)
```
epintozzi, line 3: Since `search_foods` and `count_results` are hitting the same endpoint, you could remove the `count_results` method altogether and do something like this instead so that you only have to make 1 call to the api:
```ruby
  def self.search_foods(query, page_num)
    results = FoodService.search_foods(query, page_num)
    total_hits = results[:totalHits].to_i
    food_data = results[:foods].map do |food_data|
      if food_data[:ingredients].downcase.include?(query.downcase)
        Food.new(food_data)
      end
    end.compact
    [total_hits, food_data]
  end
  ```
  
  Then in the controller you could do something like this to assign the variables:
  `@num_results, @foods = FoodFacade.search_foods(params[:search_query], 25).first(10)`

File: app/facades/food_facade.rb [hunk 2]
```diff
@@ -0,0 +1,15 @@
...
+    FoodService.search_foods(query, page_num)[:foods].map do |food_data|
+      if food_data[:ingredients].downcase.include?(query.downcase)
+        Food.new(food_data)
+      end
```
epintozzi, line 7: This is definitely above and beyond the scope of the challenge and a very neat solution to make sure we're showing more substantial data than some of the results with a lot of blanks.

epintozzi (commented):
Nice work! I know you said you ran out of time for tests, so the biggest area I'd focus on during a refactor or second look would be to add in some tests.

PR #65, opened 2022-02-23
Tim Roeder

tjroeder:
BE Mod 3 Mid-Mod assessment

### DIFF
--- a/app/controllers/foods_controller.rb
+++ b/app/controllers/foods_controller.rb
@@ -1,0 +1,17 @@
+class FoodsController < ApplicationController
+  def index
+    if params[:q].present?
+      @food_search_term = search_params
+      @food_list = FoodFacade.food_search_list(search_params)
+      @food_count = FoodFacade.food_search_list(search_params)
+    else
+      redirect_to root_path
+    end
+  end
+
+  private
+
+  def search_params
+    params.require(:q)
+  end
+end
--- a/app/controllers/foods_controller.rb
+++ b/app/controllers/foods_controller.rb
@@ -3,7 +3,7 @@ class FoodsController < ApplicationController
     if params[:q].present?
       @food_search_term = search_params
       @food_list = FoodFacade.food_search_list(search_params)
-      @food_count = FoodFacade.food_search_list(search_params)
+      @food_count = FoodFacade.food_search_count(search_params)
     else
       redirect_to root_path
     end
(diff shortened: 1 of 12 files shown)


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,17 @@
+class FoodsController < ApplicationController
+  def index
+    if params[:q].present?
```
jamisonordway, line 3: To answer the question from your submission, sad path and edge case handling were not a requirement for this assessment. I do agree that for any 'real world' project those cases should be handled and tested, so it's nice to see you've done that here.

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+      @food_search_term = search_params
+      @food_list = FoodFacade.food_search_list(search_params)
+      @food_count = FoodFacade.food_search_count(search_params)
```
jamisonordway, line 6: These lines will both eventually consume the same API endpoint. I think you could get all of this data with one API call.

app/facades/food_facade.rb
```diff
@@ -0,0 +1,13 @@
...
+    json = FoodService.search_foods(food)
+    json[:foods].map do |food_data|
+      Food.new(food_data)
+    end
+  end
+
+  def self.food_search_count(food)
+    json = FoodService.search_foods(food)
+    json[:totalHits]
+  end
```
jamisonordway, line 12: API calls can be expensive (monetarily and in terms of performance). I can see a world where we get all the data for this user story back from one method:
```
def self.food_search(food)
  json = FoodService.search_foods(food)
  food_list = json[:foods].first(10).map do |food_data|
    Food.new(food_data)
  end
  [food_list, json[:totalHits]]
end
```
Then, we could assign the values in the controller with
`@food_list, @food_count = FoodFacade.food_search(search_params)`

app/services/food_service.rb
```diff
@@ -0,0 +1,11 @@
+class FoodService
+  def self.search_foods(food)
+    response = conn.get('foods/search', { query: food, pageSize: 10 })
```
jamisonordway, line 3: It would be preferable to make `pageSize` dynamic:
```
def self.search_foods(food, page_size = 10)
  response = conn.get('foods/search', { query: food, pageSize: page_size })
  ...
end
```
I think a default argument would be great in this case, because it would default the page_size to 10 but we could always override that behavior if we wanted to.

jamisonordway (commented):
Awesome work! Your tests are robust for the scope of the assessment, and your code quality is pretty solid. 
I did leave some suggestions for reducing the amount of API calls being made. Let me know if you'd like to discuss further.

PR #66, opened 2022-02-23
Eric Mielke Mid-Mod

EMielke76:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,6 @@
+class FoodsController < ApplicationController
```
epintozzi, line 1: Nice lightweight controller! 👍🏻

app/poros/food.rb
```diff
@@ -0,0 +1,15 @@
+class Food
+  attr_reader :hits,
+              :gtin_upc,
+              :description,
+              :brand_owner,
+              :ingredients
+
+  def initialize(attributes, total_hits)
+    @hits = total_hits
```
epintozzi, line 9: I don't mind that you're storing the count on every `Food` item, however an argument against this is that is repetitive to store the same piece of unchanging data each time and takes up space in memory. It would be something more concerning in a larger app or if we were handling very large amounts of data.

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,17 @@
...
+
+<h2>Food Attributes</h2>
+<% n = 0 %>
+<% @foods.each do |food| %>
+    <section id="foods_<%= n%>">
+      <p>Description: <%=food.description%></p>
+      <p>Brand Owner: <%=food.brand_owner%></p>
+      <p>GTIN/UPC code: <%=food.gtin_upc%></p>
+      <p>Ingredients: <%=food.ingredients%></p>
+      <p>_____</p>
```
epintozzi, line 14: Nice styling touch! a `<hr>` tag will do this as well, but the line will go the whole length of the page.

epintozzi (commented):
Great testing! Very thorough and it's awesome that you were able to get vcr set up as well for this.

PR #67, opened 2022-02-23
Mid Mod Kelly A

kanderson852:


app/poros/food.rb
```diff
@@ -0,0 +1,18 @@
...
+  def initialize(data)
+    @code = data[:fgtinUpc]
+    @description = data[:description]
+    @brand = data[:brandOwner]
+    @ingredients = data[:ingredients]
+  end
+
+  def results_count
+    @results_count
+  end
```
jamisonordway, line 17: It doesn't look like this is being used- it's probably sufficient to let FoodsFacade give us this info

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,10 @@
+class FoodsController < ApplicationController
+  def index
+    @foods = FoodsFacade.top_10_results(query_params)
+    @count = FoodsFacade.foods_count(query_params)
```
jamisonordway, line 4: Both of these are hitting the same endpoint. I'll leave a comment in the facade with an example of how you could get both of these things with one API call and assign them like this:
`@foods, @count = FoodsFacade.search_results(query_params)`

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,12 @@
+class FoodsFacade
+  def self.top_10_results(string)
+    foods_list_data = FoodsService.get_foods(string)
+    foods_list_data[:foods].first(10).map do |food|
+      Food.new(food)
+    end
+  end
```
jamisonordway, line 7: I think this method could return the total count, and the top 10 food item results:
```
def search_results(string)
  foods_list_data = FoodsService.get_foods(string)
  foods = foods_list_data[:foods].first(10).map do |food|
    Food.new(food)
  end
  [foods, food_list_data[:totalHits]]
end
```
This way, you'd only make one call to the API, and you could get rid of the `foods_count` method in this facade.

app/poros/food.rb
```diff
@@ -0,0 +1,18 @@
+class Food
+  attr_reader :code,
+              :description,
+              :brand,
+              :ingredients
+
+
+  def initialize(data)
+    @code = data[:fgtinUpc]
```
jamisonordway, line 9: I think the code is missing because this should be `data[:gtinUpc]`

jamisonordway (commented):
Great work! I left some comments about ideas for refactoring the facade.
I agree that finishing the VCR setup, and maybe adding some sad path handling, would improve your tests. Overall though, your tests look good. I appreciate that you tested everything- features, poro, facade, and service.

PR #68, opened 2022-02-23
Chris Hewitt MidMod

Henchworm:
I was plagued with setup issues for the first 40 minutes, was able to complete the challenge but not as clean/well structured as I was nervous and pressed for time. Completely forgot to commit while rushing to complete challenge.


app/services/food_service.rb
```diff
@@ -0,0 +1,7 @@
+class FoodService
+
+  def self.name_search(query)
+    response = Faraday.get("https://api.nal.usda.gov/fdc/v1/foods/search?query=#{query}&dataType=&page_size=10&pageNumber=1&api_key=#{ENV['food_key']}")
```
epintozzi, line 4: I think `page_size` should've been `pageSize` and then this would have worked for you.

config/application.rb
```diff
@@ -1,4 +1,5 @@
 require_relative 'boot'
+require 'uri'
```
epintozzi, line 2: I think this might be a remnant from the error you were experiencing before removing your Gemfile.lock. If not, I'm curious about why you added it.

epintozzi (commented):
I know you said you didn't have time to refactor, but overall this really looks pretty good. I left a couple comments. Nice job with your testing. So cool to see you were able to get vcr in there in the time frame of the assessment.

PR #69, opened 2022-02-23
Robin Dittrich

Dittrir:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,8 @@
+class FoodsController < ApplicationController
+  def index
+    if params[:q].present?
```
jamisonordway, line 3: sad path/edge case handling wasn't an expectation for this assessment, but it's always good to see those handled 👍🏻

app/poros/food.rb
```diff
@@ -0,0 +1,12 @@
+class Food
+  attr_reader :total_results, :id, :description, :brand_owner, :ingredients
+
+  def initialize(data)
+    @id = data[:fdcId]
+    @description = data[:description]
+    @brand_owner = data[:brand_owner]
+    @ingredients = data[:foodNutrients].map do |nutrient|
+      nutrient[:nutrientName]
+    end
```
jamisonordway, line 10: It doesn't feel great to do a `map` within the initialize. Could this live in a method?
```
def initialize(data)
  ...
  @ingredients = ingredient_names(data[:foodNutrients])
end

def ingredient_names(nutrients)
  nutrients.map do |nutrient|
    nutrient[:nutrientName]
  end
end

File: app/views/foods/index.html.erb [hunk 3]
```diff
@@ -0,0 +1,9 @@
+<h1>Here are your results!</h1>
+
+<p>Total Results: <%= @results.count %> </p>
```
jamisonordway, line 3: `@results` has been limited to 10, meaning this will never return the true total count of search results.

app/facades/food_facade.rb
```diff
@@ -0,0 +1,9 @@
+class FoodFacade
+  def self.search_for_food(query)
+    json = FoodService.search_for_food(query)
+
+    json[:foods].map do |food_data|
+      Food.new(food_data)
+    end
+  end
```
jamisonordway, line 8: This method could return multiple values to the controller without hitting the API endpoint more than once. Specifically, the value in the local var `json` has an attribute called `totalHits` which could be used to display the total count of search results:
```
def self.search_for_food(query)
  json = FoodService.search_for_food(query)
  foods = json[:foods].first(10).map do |food_data|
    Food.new(food_data)
  end
  [json[:totalHits], foods]
end
```
You could then assign these values in the controller by doing something like
`@total_result_count, @results = FoodFacade.search_for_food(query)`

app/services/food_service.rb
```diff
@@ -0,0 +1,14 @@
+class FoodService
+  class << self
+    def conn
+      Faraday.new(url: "https://api.nal.usda.gov") do |faraday|
+        faraday.params["api_key"] = ENV['food_api_key']
+      end
+    end
+
+    def search_for_food(query)
+      response = conn.get("/fdc/v1/foods/search?query=#{query}&dataType=&pageSize=10&pageNumber=1")
```
jamisonordway, line 10: It would be better to make the `pageSize` param dynamic:
```
def search_for_food(query, page_size = 10)
  response = conn.get("/fdc/v1/foods/search?query=#{query}&dataType=&pageSize=#{page_size}&pageNumber=1")
  ...
end
```

Personally, I would prefer to take the `pageSize` param out of this call altogether, in favor of having the facade handle it.

jamisonordway (commented):
Good stuff! I like that you went above and beyond, handling a possible sad path for the user story even though it wasn't a requirement. I also think that your facade, service, and poro are mostly following conventions.
One issue is that the total count of results is not functioning as requested in the user story. This is a little bit tricky, because we need that total count but at the same time, to only display 10 results. Ideally, this data would be available from a single API call. I left some suggestions on how to approach that part of the story, but let me know if you'd like to discuss further.
It would also be good to have tests for the poro and facade.

PR #70, opened 2022-02-23
Hannah Warren

hannahkwarren:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,10 @@
+# app/controllers/foods_controller.rb
+
+require 'faraday'
```
epintozzi, line 3: Because we have `faraday` in our `Gemfile`, we don't need to manually include it here. We also don't appear to be using `faraday` in this class.

app/facades/food_facade.rb
```diff
@@ -0,0 +1,15 @@
...
+    json = FoodService.foods_containing_ingredient(query)
+    @total_hits = json[:totalHits].to_i
+  end
+
+  def self.foods_containing_ingredient(query)
+    json = FoodService.foods_containing_ingredient(query)
+    @foods = json[:foods].map do |food|
+      Food.new(food)
+    end
+  end
```
epintozzi, line 14: Since `total_hits` is calling this method, we are hitting the same api twice. Instead we can do something like this:

```ruby
  def self.foods_containing_ingredient(query)
    json = FoodService.foods_containing_ingredient(query)
    foods = json[:foods].map do |food|
      Food.new(food)
    end
    [json[:totalHits], foods]
  end
  ```
  
 And then in the controller, assign both vars at once and skip the second API call altogether:
 `@total_hits, @foods = FoodFacade.foods_containing_ingredient(params[:q])`

File: app/facades/food_facade.rb [hunk 3]
```diff
@@ -0,0 +1,15 @@
...
+
+class FoodFacade
+  def self.total_hits(query)
+    json = FoodService.foods_containing_ingredient(query)
+    @total_hits = json[:totalHits].to_i
+  end
+
+  def self.foods_containing_ingredient(query)
+    json = FoodService.foods_containing_ingredient(query)
+    @foods = json[:foods].map do |food|
```
epintozzi, line 11: We don't need foods to be an instance var here since it's not being used outside this method.

app/services/food_service.rb
```diff
@@ -0,0 +1,20 @@
...
+class FoodService 
+  def self.connection
+    Faraday.new(url: 'https://api.nal.usda.gov/fdc/v1/') do |faraday|
+      faraday.params['api_key'] = ENV['food_api_key']
+    end
+  end
+
+  def self.foods_containing_ingredient(query)
+    response = connection.get('foods/search') do |request|
+      request.params['api_key'] = ENV['food_api_key']
```
epintozzi, line 12: We can remove this line because you have already set your API key in the `connection` method above.

app/services/food_service.rb
```diff
@@ -0,0 +1,20 @@
...
+      request.params['query'] = ENV["#{query}"]
```
epintozzi, line 13: I don't think you're saving the query coming in from the controller as an environment variable anywhere, so we can use `query` on its own here:
`request.params['query'] = query`

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,10 @@
...
+
+class FoodsController < ApplicationController
+  def index
+    @total_hits = FoodFacade.total_hits(params[:query])
+    @foods = FoodFacade.foods_containing_ingredient(params[:query])
```
epintozzi, line 8: The text field label in the search form is `q`, so we don't actually have a param called `query`. If you noticed your API call wasn't returning quite what you expected, it's because there was no query param being sent with the request. If we update this to `params[:q]` it should work!

epintozzi (commented):
Nice job! You have a good foundational knowledge with the service/facade pattern in use here and your testing looks really good. You had asked how to handle missing information, and generally that is handled by making sure the view knows what to do if something isn't there. We can't control what an API will send back to us, but we can control what we do with the information we do get. That would mean either showing the field name and leaving it blank, or perhaps hiding a field name if there is no content. Sometime developers choose to filter out incomplete information altogether. How you handle it will vary on your use case. For the purposes of this assessment, it is fine to just leave it blank.

PR #71, opened 2022-02-23
Seth Perna

SethPerna:
I think I did well but not sure if I am displaying the correct info for sweet potatoes some dont have ingredients or a brand owner just seemed weird but when looking through the json response I couldnt see anything else that would make sense.


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,8 @@
+class FoodsController < ApplicationController
+  def index
+    if params[:q]
+      @food = params[:q]
```
jamisonordway, line 4: The query used in the API call should be available as part of that response (`json[:foodSearchCriteria][:query]`); I don't think you need this instance variable.

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,8 @@
+class FoodsFacade
+  def self.find_foods(food)
+    json = FoodsService.search_foods(food)
+    count = json[:totalHits]
+    foods = json[:foods].map { |food_data| Food.new(food_data, count) }
+    ten_foods = foods.shift(10)
```
jamisonordway, line 6: Very nice! I really like that you're getting 10 food results back, as well as totalHits, with just one API call.
I think it would be sufficient to return `count` and `ten_foods` from this method, rather than storing the count on each PORO. One drawback of storing the count this way is that it will take up space in local memory, but that would probably only come up as a concern on a much larger project.

jamisonordway (commented):
Great work! I'm especially stoked to see both data points being retrieved with one API call. Also, your testing is pretty thorough and you went above and beyond the assessment requirements by using VCR.
You mentioned in your submission that you'd like a deeper dive into TDDing facades/services/poros. I'd recommend simulating this experience by keeping the tests you did write, deleting those classes, and following the error messages to build them back up.

PR #72, opened 2022-02-23
Leland Curtis

LelandCurtis:


app/facades/foods_facade.rb
```diff
@@ -0,0 +1,17 @@
...
+      data = {}
+      data[:code] = food_data[:gtinUpc]
+      data[:description] = food_data[:lowercaseDescription]
+      data[:brand] = food_data[:brandOwner]
+      data[:ingredients] = food_data[:ingredients]
+      Food.new(data)
+    end
+    total_hits = json[:totalHits]
+    return [foods, total_hits]
+  end
```
jamisonordway, line 15: Really nice work on this method.
- a default argument works well here to make the API call in FoodService more dynamic and flexible.
- since you're returning both `json[:totalHits]` and 10 poros created from `json[:foods]`, there's no need for an extraneous API call.

app/services/food_service.rb
```diff
@@ -0,0 +1,14 @@
+class FoodService
+
+  def self.search(query, page_size)
+    query.gsub!(' ', "%20")
```
jamisonordway, line 4: I commented this out and everything still worked (the URI generated was `/foods?utf8=✓&q=sweet+potato&commit=Search`). According to [W3C](https://www.w3.org/Addressing/URL/uri-spec.html), `+` and `%20` are both acceptable representation of blank space in query params.

spec/rails_helper.rb
```diff
@@ -61,4 +61,9 @@
   config.filter_rails_from_backtrace!
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
+  VCR.configure do |config|
+    config.cassette_library_dir = "fixtures/vcr_cassettes"
+    config.hook_into :webmock
+    config.allow_http_connections_when_no_cassette = true
+  end
```
jamisonordway, line 68: It looks like VCR is not currently being used for any of your tests. You're probably aware of this, but for your tests to run with this VCR configuration they need a `:vcr` flag in the example definition.

jamisonordway (commented):
Great work! Overall, your code is very neat and adheres to convention.
Ideally, there would be tests for `FoodService`. However, the tests you wrote for the feature, poro, and facade are solid.

PR #73, opened 2022-02-23
Sierra Tucker

Sierra-T-9598:
Unable to get response from API

### DIFF
--- a/app/poros/food.rb
+++ b/app/poros/food.rb
@@ -1,0 +1,9 @@
+class Food
+  attr_reader :code, :description, :brand_owner, :ingredients
+  def initialize(data)
+    @code = data[:gtinUpc]
+    @description = data[:description]
+    @code = data[:brandOwner]
+    @code = data[:ingredients]
+  end
+end 
--- a/app/services/food_service.rb
+++ b/app/services/food_service.rb
@@ -13,6 +13,7 @@ class FoodService
 
     def self.get_data(url)
       response = conn.get(url)
+      require "pry"; binding.pry
       JSON.parse(response.body, symbolize_names: true)
     end
 end
--- a/app/services/food_service.rb
+++ b/app/services/food_service.rb
@@ -1,18 +1,18 @@
 class FoodService
   def self.foods_by_ingredient(ingredient)
-    search = ingredient.gsub(/\s+/, "")
-    get_data("/v1/foods/search?query=#{search}")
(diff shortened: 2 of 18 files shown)


app/services/food_service.rb
```diff
@@ -0,0 +1,19 @@
+class FoodService
+  def self.foods_by_ingredient(ingredient)
+    ingredient_search = ingredient.gsub(/\s+/, "")
+    get_data("/v1/foods/search?query=#{ingredient_search}")
```
epintozzi, line 4: Faraday can be a little picky about how we pass the url. If you remove the first forward slash, this should work and you won't get a 404 anymore:
`get_data("v1/foods/search?query=#{ingredient_search}")`

app/services/food_service.rb
```diff
@@ -0,0 +1,19 @@
+class FoodService
+  def self.foods_by_ingredient(ingredient)
+    ingredient_search = ingredient.gsub(/\s+/, "")
```
epintozzi, line 3: This shouldn't be necessary. Faraday will do this for us. You can pass `ingredient` directly to the `get_data` method on line 4 and it should work!

app/poros/food.rb
```diff
@@ -0,0 +1,9 @@
+class Food
+  attr_reader :code, :description, :brand_owner, :ingredients
+  def initialize(data)
+    @code = data[:gtinUpc]
+    @description = data[:description]
+    @code = data[:brandOwner]
+    @code = data[:ingredients]
```
epintozzi, line 7: Watch your instance variables. This was probably just a copy/paste oversight but these will compete with each other as the same ivar can't have multiple values.

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,9 @@
+<p>Total Results: <%= @foods.count%> </p>
+
+<p>Showing 10 results: </p>
+<% @foods.each do |food| %>
+  <p>GTIN/UPC:<%= food.code %></p>
+  <p>Description:<%= food.description %></p>
+  <p>Brand Owner:<%= food.brand_owner %></p>
+  <p>Ingredients:<%= food.ingredients %></p>
```
epintozzi, line 8: I bet these will work once you update your poro to have unique instance variables

epintozzi (commented):
Nice job! You were soooo close with the API consumption. Just one tiny change and it'll be working for you. I'd like to see a little more testing, but overall you showed great use of the service/facade pattern.

PR #74, opened 2022-02-23
Croix Kimmel mid mod

croixk:


Gemfile
```diff
@@ -17,7 +17,7 @@ gem 'jbuilder', '~> 2.5'
 gem 'bootsnap'
 gem 'faraday'
 gem 'figaro'
-
+gem 'dotenv-rails'
```
epintozzi, line 20: 👀  I love seeing when y'all try out tools other than the ones we talk about in class. Nice!

app/facades/food_facade.rb
```diff
@@ -0,0 +1,10 @@
+class FoodFacade
+  def self.foods(food)
+    json = FoodService.search_foods(food)
+    count = json[:totalHits]
+    json[:foods].map do |food_data|
+      Food.new(food_data, count)
```
epintozzi, line 6: Neat choice to include the count here! A counter argument to this approach might be that you are storing count on every `Food` item which ultimately takes up space in memory. It's not a huge lift and I don't mind, but for a large scale app or if you were handling a lot of data, it could be something that would come up.

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,6 @@
+class FoodsController < ApplicationController
+  def index
+    @food = params[:q]
```
epintozzi, line 3: I don't think you're using this instance var anywhere except on line 4 so you could probably use a local variable instead, or even remove this line altogether and pass the param directly in line 4 like this:
`@foods = FoodFacade.foods(params[:q])`

epintozzi (commented):
Overall you were definitely on the right track. It looks like you weren't hitting quite the right endpoint because it didn't look like I was getting search results for my search term, however I think that's probably just a tweak to the url you are going to with Faraday. Otherwise, the code is pretty well organized. I'd add in more testing for your service, poro, and facade but this was a great start.

PR #75, opened 2022-02-23
Eldridge Turambi Mid Mod

Eldridge-Turambi:
I did not finish. Got stuck in the beginning with some unknown errors and then struggled getting postman to work. Got as far as I could in making poro and facade folders.


epintozzi (commented):
This is a great starting point! I'd love to see a finished version sometime to give more in depth feedback.

PR #76, opened 2022-02-23
Joel Grant

joel-grant:


app/views/foods/index.html.erb
```diff
@@ -0,0 +1,12 @@
+<h1>Total Search Results: <%= @count %></h1>
+
+<p><h3>Foods Containing Sweet Potatoes:</h3></p>
```
jamisonordway, line 3: I'm pretty sure this was meant to be dynamic?

app/services/food_service.rb
```diff
@@ -0,0 +1,8 @@
+class FoodService
+  def self.search_all_items(search)
+    search = ERB::Util.url_encode search
+    response = Faraday.get("https://api.nal.usda.gov/fdc/v1/foods/search?api_key=#{ENV['API_KEY']}&query=#{search}")
+    # response = Faraday.get("https://api.nal.usda.gov/fdc/v1/foods/search?api_key=#{ENV['API_KEY']}&query=#{search}&pageSize=10")
```
jamisonordway, line 5: I agree that the FoodFacade should be responsible for limiting results 👍🏻

jamisonordway (commented):
Great work overall! 
I think handling the food count as well as a max of 10 search results is tricky. Your solution is good, though, and without mocks and stubs I would expect it to slow down your tests since there are so many results.
I would not limit the search results by 10 in the FoodService for a couple of reasons. 
1) We have a facade which exists to reformat/reorganize the data we're getting from the API, so putting that kind of logic in a service feels like an SRP violation. 
2) limiting the results to 10 in the service might create a situation where we're always getting '10' back for a count of foods, or force us to make an additional call just to get the total count of results.

I would also test FoodService and the PORO, but your tests for the facade and index are pretty thorough.

PR #77, opened 2022-02-23
Parker Lockhart

ParkerLockhart:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,6 @@
+class FoodsController < ApplicationController
+  def index
+    @results = FoodFacade.foods_search(params[:q])
+    @info = FoodFacade.search_info(params[:q])
```
jamisonordway, line 4: Both of these lines are ultimately pinging the same API endpoint. I'll leave some suggestions in other files about how we might be able to ping the API just once and still have both data points available for assignment here.

app/facades/food_facade.rb
```diff
@@ -0,0 +1,12 @@
+class FoodFacade
+
+  def self.foods_search(query)
+    FoodService.foods_search(query)[:foods].map do |data|
+      Food.new(data)
+    end
+  end
```
jamisonordway, line 7: I can see a world where this method does more heavy lifting, returning `totalHits` as well as the Food POROs:
```
def self.foods_search(query)
  food_data = FoodService.foods_search(query)
  foods = food_data.first(10).map do |data|
    Food.new(data)
  end
  [food_data[:totalHits], foods]
end
```

With this implementation, you would not need the `SearchInfo` PORO or the `search_info` method below. I don't necessarily think that having a `SearchInfo` PORO is a poor design choice, but there is the drawback of hitting the same endpoint more times than is needed for the requested behavior.

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,9 @@
+<h1>RESULTS FOR <%= @info.query %></h1>
+Total Results: <%= @info.total_hits %><br><br>
+
+<% @results.first(10).each do |result| %>
```
jamisonordway, line 4: You could limit `@results` to 10 earlier, maybe in the facade, to avoid passing a bunch of data to the view that will never be displayed.

jamisonordway (commented):
Really awesome work, especially on your tests 🎉  It's great that you were so specific in them and that you were able to use VCR. I'd recommend adding the css to get the feature tests passing, like you mentioned in your submission.
There are a couple of opportunities to improve code quality, which I left comments about. Generally, try to leverage facades to manipulate the data you get back from the services, and look for ways to ping API endpoints as seldom as possible.

PR #78, opened 2022-02-23
Kerri Hoffmann

kbhoffmann:


app/services/food_service.rb
```diff
@@ -0,0 +1,10 @@
+class FoodService
+  def self.conn
+    Faraday.new(url: "https://api.nal.usda.gov")
+  end
+
+  def self.get_foods_by_type(food_type)
+    response = conn.get("/fdc/v1/foods/search?query=#{food_type}&pageSize=50&api_key=#{ENV["food_api_key"]}")
```
jamisonordway, line 7: This imposes a `pageSize` on every API call. It would be more conventional to make this param dynamic:
```
def self.get_foods_by_type(food_type, page_size = 10)
  response = conn.get("/fdc/v1/foods/search?query=#{food_type}&pageSize=#{page_size}&api_key=#{ENV["food_api_key"]}")
  data = JSON.parse(response.body, symbolize_names: true)
end
```
In this case, though, I might suggest returning all results and having the facade handle the limiting logic instead. That way, we could get a count of total results and the 10 foods to show, all with one API call. I'll leave a comment in the facade with an example of what that might look like.

app/facades/food_facade.rb
```diff
@@ -0,0 +1,8 @@
+class FoodFacade
+  def self.searched_food(food_type)
+    foods = FoodService.get_foods_by_type(food_type)[:foods]
+    foods.map do |food_data|
+      Food.new(food_data)
+    end
+  end
```
jamisonordway, line 7: We could use this method to return both data points needed for the user story:
```
def self.searched_food(food_type)
  food_data = FoodService.get_foods_by_type(food_type)
  foods = food_data[:foods].first(10).map do |food|
    Food.new(food)
  end
  [food_data[:totalHits], foods]
end
```
The controller would assign instance vars like so: 
`@result_count, @foods = FoodFacade.searched_food(params[:q])`

jamisonordway (commented):
While the user story isn't complete, the functionality here looks good! Your facade, service, and poro are mostly conventional. 
In your submission you mentioned plans to use `totalHits` from the API call and add some logic to limit search results to 10. It sounds like you were headed in the right direction, and I've left some comments on how I might approach that part of the user story.

PR #79, opened 2022-02-23
Arnaldo Aparicio

arnaldoaparicio:


app/views/foods/index.html.erb
```diff
@@ -0,0 +1,9 @@
+<h1>Results</h1>
+
+<p>Total Items: <%= @food_results.count %> </p>
```
epintozzi, line 3: Ideally this would be a count of all foods containing sweet potatoes (which is something like 30,000). Since your API request is imposing a page size limit of 10 this will always be 10. I would probably remove the limit of 10 on the API request in the service to get the count and then take the first 10 items and assign them to POROs for the view.

epintozzi (commented):
Overall great work. I think the tricky part of this one is the total count of foods. We have to make technical decisions and trade offs on how to best handle that, and it feels inefficient to request all foods just for a count when we can limit to the 10 we want to display.

Good work on your tests! I'd probably also test the facade but the rest of the tests are pretty comprehensive 👍🏻

PR #80, opened 2022-02-23
Karan Mehta

karanm645:
I was getting stuck in connecting the facade to the services. For some reason it wasn’t recognizing it and would love to know how I could’ve gotten past that. I was able to set up the connection and run the endpoints in Postman. I didn’t get too far because of this hiccup, but it was a good assessment to see where I’m at and where I need to improve!


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,5 @@
+class FoodsController < ApplicationController
+  def index
+    @food = FoodService.all_food
```
epintozzi, line 3: In order to properly make our query, we need to pass the user's query param from the controller to the service. We could do it like this:
`@food = FoodService.all_food(params[:q])`

app/services/food_service.rb
```diff
@@ -0,0 +1,13 @@
+require 'faraday'
```
epintozzi, line 1: Since faraday is in our `Gemfile` we shouldn't need to include it here.

app/services/food_service.rb
```diff
@@ -0,0 +1,13 @@
...
+class FoodService
+  def self.connection
+    Faraday.new(url: "https://api.nal.usda.gov/fdc/v1/foods/") do |f|
+      f.params['api_key'] = 'Ze3akdRIKt8TZiohIkLEfF5va0Q5WHAshPSFaHkl'
```
epintozzi, line 5: We don't want to commit our API keys. I recommend using a tool like `figaro` to help manage your keys. Then, you could call it as an environment variable here instead, like this:
` f.params['api_key'] = ENV['food_api_key']`

app/services/food_service.rb
```diff
@@ -0,0 +1,13 @@
...
+    end
+  end 
+
+  def self.all_food
+    response = connection.get("list")
```
epintozzi, line 10: We should be passing the user's query param to this method from the controller and our argument here should be the rest of the path, so something like:
`response = connection.get("foods/search?query=#{query}")`

config/routes.rb
```diff
@@ -1,3 +1,4 @@
 Rails.application.routes.draw do
   root "welcome#index"
+  resources :foods
```
epintozzi, line 3: We're only creating an `index` right now so we should limit to only the index action: `resources :foods, only: :index`

PR #82, opened 2022-02-23
Josh Walsh

jaw772:


app/services/food_service.rb
```diff
@@ -0,0 +1,10 @@
+class FoodService
+  def food_search(query)
+    get_url("/search?api_key=#{ENV['food_api_key']}&query=#{query}&pageSize=10&pageNumber=1")
```
jamisonordway, line 3: It would be a good idea to make the facade handle the limiting logic instead of hard coding a `pageSize` parameter into this API call. It's possible to make one API call for the 10 POROs and the total result count at the same time.
Something like:
```
# in FoodService
def food_search(query)
  get_url("/search?api_key=#{ENV['food_api_key']}&query=#{query}
end

# in FoodFacade

def food_search(query)
  results = service.food_search(query)[:foods].map(&:food)
  foods = results.take(10).map { |food| FoodSearch.new(food) }
  {total_result_count: results[:totalHits], foods: foods}
end
```

jamisonordway (commented):
One piece of functionality is missing from the user story: displaying a total count of search results. Other than that, this is pretty solid.

To address a couple of comments from your submission:
- edge case & sad path testing, along with VCR/Webmock, were not requirements for this mid mod. I do agree that they'd be great things to add!
- I recommend [these docs](https://docs.github.com/en/get-started/getting-started-with-git/managing-remote-repositories) for any situation where you need to add or remove a remote.

PR #84, opened 2022-02-23
Brad Breiten - Mid Mod Submission

jbreit88:


app/poros/hits.rb
```diff
@@ -0,0 +1,7 @@
+class Hits
```
epintozzi, line 1: Your submission notes said to ignore this so ignore I shall :)

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,9 @@
+class FoodsController < ApplicationController
+
+  def index
+    @facade = FoodDataFacade.new
+    @total_hits = @facade.total_hits(params[:q])
+    @foods = @facade.search_by_ingredient(params[:q]).first(10)
```
epintozzi, line 6: Ultimately these 2 lines are hitting the same endpoint. You could clean this up by doing something like this in your service:

```ruby
  def search_by_ingredient(query)
    food_data = service.search_by_ingredient(query)
    foods = food_data[:foods].map do |data|
      Food.new(data)
    end
    [foods, food_data[:totalHits]]
  end
  ```
  
  And then in your controller assign like this:
  `@foods, @total_hits = @facade.search_by_ingredient(params[:q]).first(10)`
  
Then you can get rid of the `total_hits` method in the facade altogether as well as make just one call to the API.

File: app/facades/food_data_facade.rb [hunk 3]
```diff
@@ -0,0 +1,18 @@
+class FoodDataFacade
+  attr_reader :service
+
+  def initialize
+    @service = FoodDataService.new
+  end
+
+  def total_hits(query)
+    service.search_by_ingredient(query)[:totalHits]
+  end
```
epintozzi, line 10: See comment above regarding how we can get rid of this method.

spec/facades/food_data_facade_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'rails_helper'
+
+RSpec.describe FoodDataFacade do
```
epintozzi, line 3: Good job testing here. I like that you are testing that the Facade is returning an array and the class of each item in the array and nothing more. The facade doesn't care about more than that and it can be easy to overtest this stuff.

spec/welcome/index_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      scenario 'On /foods I should see attributes for each of the ten foods' do
+        within "#food-1" do
+          expect(page).to have_content('UPC')
+          expect(page).to have_content('Description')
+          expect(page).to have_content('Brand')
+          expect(page).to have_content('Ingredients')
+        end
+      end
+    end
+  end
```
epintozzi, line 48: Super thorough testing here! This whole spec is great, but you've really gone above and beyond to make sure you have everything in your view that you are expecting without anything more.

epintozzi (commented):
Great work! I think your testing is very solid and this is what we expect for a midmod. If you wanted to be super thorough you could add in edge cases (what happens if a user leaves the search term blank for example) but that it not expected for your mid mod.

PR #85, opened 2022-02-23
Devin 

devinpile:
finished the core of it in 1hr 15mins and spent the rest of the time reading docs to learn more about the API and help me debug... 
i feel like i get it... for the most part... but when it doesn't work, and i think that it should, it makes my confidence shaky

### DIFF
--- a/app/facades/food_facade.rb
+++ b/app/facades/food_facade.rb
@@ -12,4 +12,4 @@ class FoodFacade
       data[:totalHits]
     end
   end
-end 
+end
--- a/app/facades/food_facade.rb
+++ b/app/facades/food_facade.rb
@@ -1,0 +1,15 @@
+class FoodFacade
+  class << self
+    def food_search(food)
+      data = FoodService.find_foods(food)
+      data[:foods].take(10).map do |food_data|
+        Food.new(food_data)
+      end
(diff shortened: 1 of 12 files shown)


app/facades/food_facade.rb
```diff
@@ -0,0 +1,15 @@
+class FoodFacade
+  class << self
+    def food_search(food)
+      data = FoodService.find_foods(food)
+      data[:foods].take(10).map do |food_data|
+        Food.new(food_data)
+      end
+    end
```
epintozzi, line 8: Since the `result_count` method is hitting the same endpoint as the `food_search` method, we can actually consolidate this a bit and make our requests more efficient by taking all the info we need the first time. We could do something like this and remove the `result_count` method:

```ruby
  def food_search(food)
      foods = data[:foods].take(10).map do |food_data|
        Food.new(food_data)
      end
      [data[:totalHits], foods]
    end
```

Then in the controller, you can do:
`@result_count, @foods = FoodFacade.food_search(params[:q])`

epintozzi (commented):
This looks great! Very nicely written. I left one refactor suggestion. Your tests are super thorough as well. Awesome work!

PR #90, opened 2022-05-26
Joseph Downs

josephdowns:


app/facades/food_facade.rb
```diff
@@ -0,0 +1,15 @@
+class FoodFacade
+  class << self
+    def search_for_foods(food)
+      food_data = FoodService.get_data("search?query=#{food}")
+      foods = food_data[:foods].map do |result|
+        Food.new(result)
+      end
+    end
```
jamisonordway, line 8: This method could return the first 10 foods as well as the total count of results. That way, you wouldn't be hitting the API endpoint twice for this story.
```
def search_for_foods(food)
  food_data = FoodService.get_data("search?query=#{food}")
  total_results = food_data[:totalHits]
  
  # only create Food objects out of the first 10
  foods = food_data[:foods].first(10).map do |result|
    Food.new(result)
  end
  # return an array of values to the controller
  [foods, total_results]
end
```
```
# in FoodsController
def index
  @foods, @total_results = FoodFacade.search_for_foods(params[:q])
end
```

PR #92, opened 2022-05-27
Kim Guo

kg-byte:
Completed user story with testing; to ensure consistency of data, the following params were added to the api call:

	faraday.params['pageSize'] = 10
	faraday.params['sortBy'] = 'dataType.keyword'
	faraday.params['sortOrder'] = 'asc'


kg-byte:
I realized that I missed the 'total number of hits' requirementm which can be handled by the FoodsFacade.search_ingredient(ingredient) method and a helper search_data method. This is what I would have done:
```
class FoodsFacade
  def self.search_ingredient(ingredient)
    search_data(ingredient)
    {total_hits: search_data[:total_hits], foods: search_data[:foods].map { |datum| Food.new(datum)} }
  end
  
   def search_data(ingredient)
      search_data ||= FoodsService.search_ingredient(ingredient)
   end
end
```

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,28 @@
+  <h1>Search results</h1>
+
+  <table class="table table-hover">
```
juliet-e, line 3: I appreciate the formatting done here to present the data in a more readable way!

juliet-e: It is just missing the "total foods" part of the user story

spec/facades/foods_facade_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'rails_helper'
+
+RSpec.describe FoodsFacade do
+  it 'gets data for 10 foods based on keyword search' do
+    data = JSON.parse(File.read('spec/fixtures/cheese_data.json'), symbolize_names: true)
+    allow(FoodsService).to receive(:search_ingredient).and_return(data)
```
juliet-e, line 6: Nice work isolating this into a pure unit test with stubbed data! Looks great

spec/services/foods_service_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'rails_helper'
+
+RSpec.describe FoodsService do
+  it 'gets data for 10 foods based on keyword search' do
+    data = JSON.parse(File.read('spec/fixtures/cheese_data.json'), symbolize_names: true)
+    allow(FoodsService).to receive(:search_ingredient).and_return(data)
```
juliet-e: When we are testing the service itself, we *don't* want to mock its return value, as that will make our test a bit useless (we're testing that the mock works rather than the actual logic in our service). So here, you'd want to use webmock or vcr because we *do* want to make sure that our service is receiving data from the external API, parsing the json, etc. Please reach out if you have any questions about this comment -- happy to chat.

PR #93, opened 2022-05-27
Johnny Bowman

johnny-bowman:


app/facades/food_facade.rb
```diff
@@ -0,0 +1,9 @@
+class FoodFacade
+  def self.search_by_ingredient(ingredient)
+    json = FoodService.search_by_ingredient(ingredient)
+
+    json[:foods][0..9].map do |food_data|
+      Food.new(food_data)
+    end
```
jamisonordway, line 7: Nice job limiting the number of PORO instances being created here. This is definitely something the facade should be responsible for, and I like that this is not creating more objects than we would ever want to display.

app/services/base_service.rb
```diff
@@ -0,0 +1,11 @@
...
+  private
+
+  def self.conn
+    Faraday.new('https://api.nal.usda.gov')
+  end
+
+  def self.get_json(response)
+    JSON.parse(response.body, symbolize_names: true)
+  end
+end
```
jamisonordway, line 11: Good use case for inheritance. 👍🏻

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,7 @@
+class FoodsController < ApplicationController
+  def index
+    @results_count = FoodService.search_by_ingredient(params[:q])[:totalHits]
```
jamisonordway, line 3: This is the same endpoint being hit for the 10 food objects. Ideally, we would only make 1 API call for both of these data points.

PR #94, opened 2022-05-27
Alicia Watt

AliciaWatt:


app/facades/foods_facade.rb
```diff
@@ -0,0 +1,15 @@
...
+      data = {}
+      data[:code] = food_data[:gtinUpc]
+      data[:description] = food_data[:lowercaseDescription]
+      data[:brand] = food_data[:brandOwner]
+      data[:ingredients] = food_data[:ingredients]
+      Food.new(data)
+    end
+    total_hits = json[:totalHits]
+    [foods, total_hits]
+  end
```
jamisonordway, line 14: Awesome! I really like that the facade is handling these responsibilities, and that you're only calling the endpoint once for both of these data points. 
I do wonder if lines 5-9 are necessary, or if the PORO initialize could accept the `food_data` block variable, but I don't have as strong of a preference about that.

PR #96, opened 2022-05-27
Will Medders

wmedders21:
MidMod submission
Screen Shot 2022-05-27 at 12 37 08 PM

### DIFF
--- a/app/facades/foods_facade.rb
+++ b/app/facades/foods_facade.rb
@@ -1,0 +1,9 @@
+class FoodsFacade
+
+  def self.ten_foods(query)
+    foods_list = FoodDataService.parsed_search_response(query)[:foods].take(10)
+    foods_list.map do |food|
+      Food.new(food)
+    end
+  end
+end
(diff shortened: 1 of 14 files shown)


config/application.yml
```diff
@@ -0,0 +1 @@
+api_key: 'iXuv3TmRRccB1MSd0QHHDcRdE04vcF7FhXhEqh5t'
```
juliet-e, line 1: careful not to check this in to Github! Figaro install should automatically add this to your .gitignore

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,9 @@
+class FoodsFacade
+
+  def self.ten_foods(query)
+    foods_list = FoodDataService.parsed_search_response(query)[:foods].take(10)
```
juliet-e, line 4: Is there a way to limit the number of results to 10 without this ruby logic?

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,8 @@
+<h1>Foods containing <%= params[:q] %></h1>
```
juliet-e, line 1: Just missing total hits for a given food. How could you retrieve that data?

PR #97, opened 2022-05-27
Alex Psitos whats in my food PR

psitosam:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,10 @@
+class FoodsController < ApplicationController
+
+
+  def index
+    if params[:q]
+      @food = params[:q]
```
jamisonordway, line 6: I don't think this needs to be an instance variable. The results from the API should provide a food name that you could send to the view.

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,8 @@
+class FoodsFacade
+  def self.find_foods(food)
+    json = FoodsService.search_foods(food)
+    count = json[:totalHits]
+    foods = json[:foods].map { |food_data| Food.new(food_data, count) }
+    ten_foods = foods.shift(10)
```
jamisonordway, line 6: This will potentially create hundreds of Food PORO objects which will never be returned to the controller. You could limit the number of results earlier:
```
foods = json[:foods].first(10).map do |food_data|
  Food.new(food_data, count)
end
```

spec/facades/foods_facade_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'rails_helper'
+
+RSpec.describe FoodsFacade do
+  it '#find_foods', :vcr do
```
jamisonordway, line 4: yeah VCR! 🎉

PR #98, opened 2022-05-27
Zel Imbriaco Submission

zel-imbriaco:
Using the ingredients search produces results with no gtinUpc, brandOwner, or ingredients field.

### DIFF
--- a/app/facades/food_facade.rb
+++ b/app/facades/food_facade.rb
@@ -1,0 +1,3 @@
+class FoodFacade
+
+end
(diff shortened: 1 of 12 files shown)


app/facades/food_facade.rb
```diff
@@ -0,0 +1,18 @@
...
+      food_data = {
+        gtinUpc: food[:gtinUpc],
+        description: food[:description],
+        brandOwner: food[:brandOwner],
+        ingredients: food[:ingredients]
+      }
+      results[:foods] << Food.new(food_data)
+    end
+    results
+  end
```
jamisonordway, line 17: Awesome 🎉  I really like that you've returned all data points for the user story with a single call to the API.

PR #99, opened 2022-05-27
Brylan Gannon 

B-gann21:


app/facades/food_facade.rb
```diff
@@ -0,0 +1,11 @@
+class FoodFacade
+  def self.total_results(query)
+    FoodService.get_foods(query)[:totalHits]
+  end
+
+  def self.find_foods(query)
+    body = FoodService.get_foods(query)
+    
+    body[:foods][0..9].map { |data| Food.new(data) }
+  end
```
jamisonordway, line 10: 👏🏻  Great job limiting the number of foods to be created. That's definitely a facade responsibility, and it's better to limit that collection **before** using it to create PORO instances.
💡 You could make this more efficient by returning `total_hits` and `foods` from the same method. Since these are consuming the same API endpoint, we want to avoid hitting it more than necessary. 
```
def self.find_foods(query)
  results = FoodService.get_foods(query)
  total_hits = results[:totalHits]
  foods = results[:foods][0..9].map { |data| Food.new(data) }
  [total_hits, foods]
end
```

PR #100, opened 2022-05-27
Luke Pascale

enalihai:
I didnt do any tests, and got hung up on my postman route. Otherwise I think I knew what I was doing, just needed more practice to be faster.

### DIFF
--- a/app/facades/food_facade.rb
+++ b/app/facades/food_facade.rb
@@ -1,0 +1,13 @@
+class FoodFacade
+  def self.display_count_foods_returned
+    # binding.pry
+    json = FoodService.count_foods_returned[:totalHits]
+
+    # json[:totalPages]
+
+    # binding.pry
+    # json[:results].map do |food_data|
+    #   Food.new(food_data)
+    # end
+  end
+end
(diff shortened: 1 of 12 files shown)


app/facades/food_facade.rb
```diff
@@ -0,0 +1,13 @@
...
+    # binding.pry
+    json = FoodService.count_foods_returned[:totalHits]
+
+    # json[:totalPages]
+
+    # binding.pry
+    # json[:results].map do |food_data|
+    #   Food.new(food_data)
+    # end
+  end
```
jamisonordway, line 12: I think a method similar to this could return the total count *and* the first 10 foods as PORO instances:
```
def self.food_search_results(query)
  json = FoodService.count_foods_returned
  total_hits = json[:totalHits]
  
  foods = json[:results].first(10).map do |food_data|
    Food.new(food_data)
  end
  [total_hits, foods]
end
```
This way, you'd only make one call to the API, and could avoid creating Food PORO instances that would never be displayed.

app/services/food_service.rb
```diff
@@ -0,0 +1,23 @@
+class FoodService
+  def self.conn
+    Faraday.new("https://api.nal.usda.gov")
+  end
```
jamisonordway, line 4: Instead of configuring the API key param in multiple places, you could set that up here. Something like:
```
def self.conn
  Faraday.new(url: "https://api.nal.usda.gov", params: {api_key: ENV["fdc_api_key"]})
end
```

PR #103, opened 2022-05-27
Tyler Ross MidMod Mod3

Tross0208:
Currently stuck trying to create Food poro from facade.


app/controllers/poros/food.rb
```diff
@@ -0,0 +1,12 @@
+class Food
```
juliet-e, line 1: It will be hard for Rails to find this class when poros is in controllers. If you bring the poros directory into the app directory, you shouldn't run into the errors you were seeing

PR #104, opened 2022-05-27
Chris Romero

CLRM1:


app/services/food_data_service.rb
```diff
@@ -0,0 +1,13 @@
...
+    response = conn.get("foods/search?query=#{params[:q]}")
+    @data = JSON.parse(response.body, symbolize_names: true)
+  end
+
+  def conn
+    connection = Faraday.new(url: "https://api.nal.usda.gov/fdc/v1/") do |f|
+      f.headers["X-API-KEY"] = ENV['food_data_api_key']
+    end
+  end
+end
```
jamisonordway: I'd say you're on the right track with this refactor. One note about this class as well as `FoodDataFacade`: typically we would expect these to use class methods, so that there isn't a need to instantiate them each time we want to make an API call.

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,13 @@
+<h1>Search Results</h1>
+
+<h2>Total results: <%= @data[:totalHits]  %> </h2>
+
+<h2>Foods containing <%= @data[:foodSearchCriteria][:query] %> </h2>
+
+<% @results[0..9].each do |food_item| %>
```
jamisonordway, line 7: limiting the number of food results feels more like a Facade responsibility.

PR #105, opened 2022-05-27
Alex Randolph

alexGrandolph:


app/views/search/foods.html.erb
```diff
@@ -0,0 +1,15 @@
+<h1>Food Result Page</h1>
+
+<h3>Results:<h3>
+
+Results Count: <%= @foods.count %><br>
```
juliet-e, line 5: There is an even easier way to get this data if you check out the response from that USDA endpoint

spec/rails_helper.rb
```diff
@@ -62,3 +62,9 @@
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
 end
+VCR.configure do |config|
+  config.cassette_library_dir = "spec/fixtures/vcr_cassettes"
+  config.hook_into :webmock
+  config.filter_sensitive_data('movie_db_key') { ENV['e2d1402726608719026ca6424f229bcf'] }
```
juliet-e, line 68: Careful here! Please use `ENV['food_data_key']` instead of your actual API key!

PR #106, opened 2022-05-27
Andrew Speth

aspeth:
I got most of the functionality built, couldn't figure out how to get the number of results to be visible in the controller or view after refactoring to facade/service/poro setup. I know I could do it by moving some of the logic in my facade into the controller but that didn't seem like the right answer, I'd love feedback on how to make that happen.


app/facades/foods_facade.rb
```diff
@@ -0,0 +1,18 @@
+class FoodsFacade
+  def self.service
+    @_service ||= FoodsService.new
```
juliet-e, line 3: Typically, service methods are class methods so that you don't have to instantiate them like this.

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  def self.search_by_keyword(word)
+    filtered_results = []
+    
+    service.search_by_keyword(word)[:foods].each do |food|
+      if food[:ingredients].downcase.include?(word)
```
juliet-e, line 10: Excellent filtering!

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,18 @@
...
+        filtered_results << food
+      end
+    end
+    filtered_results.first(10).map do |food|
```
juliet-e, line 14: Because you added the filtering (which was actually an extension - we caused some confusion there - so nice work!) I think it makes sense to grab the first 10. However, if we could assume that all food hits would come with ingredients that are non-empty, we'd prefer to limit the result set through the API call with a query param (check out the docs to explore how that would work). But for this case, this limiting makes total sense.

app/services/foods_service.rb
```diff
@@ -0,0 +1,11 @@
+class FoodsService
+  def get_url(url)
+    conn = Faraday.new("https://api.nal.usda.gov/fdc/v1/foods/")
+    response = conn.get(url)
+    JSON.parse(response.body, symbolize_names: true)
+  end
+
+  def search_by_keyword(word)
+    get_url("search?query=#{word}&dataType=&sortBy=dataType.keyword&sortOrder=asc&api_key=#{ENV['api_key']}")
```
juliet-e, line 9: You can also consider adding the API key param to the connection so it's always there

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,18 @@
+class FoodsFacade
```
juliet-e, line 1: To answer your question about total hits: What if your facade returned data for both the 10 foods AND total hits. This could look a couple different ways -- a parent PORO that stores all the foods and the total hits, or even just a simple hash you make on the spot to return both back to the controller.

spec/rails_helper.rb
```diff
@@ -62,3 +62,8 @@
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
 end
+VCR.configure do |config|
+  config.cassette_library_dir = 'spec/fixtures/vcr_cassettes'
+  config.hook_into :webmock
+  config.configure_rspec_metadata!
```
juliet-e, line 68: Just a reminder for the future -- add configuration to obscure your API key in the url so that your cassettes don't expose your key. For this assessment it's totally fine. Just good security hygiene in general to always include that.

PR #107, opened 2022-05-27
Ian Ross

ross-ian28:


app/facades/food_facade.rb
```diff
@@ -0,0 +1,5 @@
+class FoodFacade
+  def self.get_foods(params)
+    FoodService.get_searched_food(params).map { |data| Food.new(data) }
```
juliet-e, line 3: Make sure you take a look at the response structure in Postman or through a pry. This response actually wasn't an array at the top level. You had to instead go into the [:foods] key and then map over that in order to get the correct data.

PR #108, opened 2022-05-27
Phillip Kamps

phillipkamps:
Completed most(?) of challenge, then saw we needed to only show results with ingredients...
Was I on the right track? See the `foods_facade.rb` `self.food_results` method comments for where I was trying to go.


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,6 @@
+class FoodsController < ApplicationController
+  def index
+    @total_hits = FoodsFacade.total_hits(params[:q])
+    @food_results = FoodsFacade.food_results(params[:q])
```
juliet-e, line 4: Can you think of a way to provide both hits and food data with just *one* API call?

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,12 @@
+class FoodsFacade
+  def self.total_hits(params)
+    results = FoodsService.get_search(params)
+    results[:totalHits]
+  end
+
+  def self.food_results(params)
+    FoodsService.get_search(params)[:foods][0..9].map { |data| Food.new(data) }
```
juliet-e, line 8: How can you limit results to 10 without having to add this ruby logic?

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,21 @@
+require "rails_helper"
+
+RSpec.describe "Foods search results page" do
```
juliet-e, line 3: I appreciate the readability of these tests!

spec/services/foods_service_spec.rb
```diff
@@ -0,0 +1,11 @@
+require "rails_helper"
+
+RSpec.describe "Foods Service" do
```
juliet-e, line 3: Nice work testing the service! Remember to also include Facade and PORO tests.

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,12 @@
...
+    # results = FoodsService.get_search(params)
+    # wip = results[:foods].find { |result| result[:ingredients].include?(params) }.limit(10)
```
juliet-e, line 10: Great start for filtering out empty ingredients! Apologies for the confusion -- this wasn't actually necessary and also would've been an extension, but you've got the right idea with this filter functionality.

PR #109, opened 2022-05-27
Reuben Davison 

Reuben-Davison:


app/facades/foods_facade.rb
```diff
@@ -0,0 +1,15 @@
...
+            Food.new(data)
+        end
+    end 
+
+    def get_foods_data(search)
+        service.get_foods(search)
+    end
+    
+    def service 
+        FoodService.new
```
jamisonordway, line 13: Generally, we would expect Services and Facades to use class methods, so that instances like this are not necessary.

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,12 @@
+class FoodsController < ApplicationController 
+    def index 
+        if params[:q] != ""
+            @foods = FoodsFacade.new.get_foods_search(params[:q])
+            @search_count = FoodsFacade.new.get_foods_data(params[:q])[:totalHits]
```
jamisonordway, line 5: This is consuming the same API endpoint twice. We could get both of these data points with a single call (see my comment in FoodsFacade)

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,15 @@
+class FoodsFacade 
+    def get_foods_search(search)  
+        get_foods_data(search)[:foods].map do |data|
+            Food.new(data)
+        end
+    end
```
jamisonordway, line 6: The facade would be a great place to handle the responsibility of only returning 10 Food results. We can also get both the `total_hits` with this call, something like: 
```
def get_foods_search(search)
  results = get_foods_data(search)
  total_hits = results[:totalHits]
  foods = results[:foods].first(10).map do |data|
    Food.new(data)
  end
  [total_hits, foods]
end
```
This way, you wouldn't be making extraneous calls to the API, and you wouldn't create unused instances of Food.

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,14 @@
+<h1>Search Results for <%= params[:q] %> </h1>
+
+
+<h3>Total search results: <%= @search_count %></h3>
+
+<% @foods[0..9].each do |food| %>
```
jamisonordway, line 6: As I mentioned in another comment, limiting the number of foods returned seems more like a Facade responsibility.

PR #110, opened 2022-05-27
Susan Boyd

sueboyd922:


app/poros/food_info.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  def with_ingredient(search)
+    # require "pry"; binding.pry
+    @foods.select do |food|
+      unless food[:ingredients].nil?
+        food[:ingredients].downcase.include?(search.downcase)
+      end
+    end[0..9]
+  end
```
jamisonordway, line 16: We could have been more clear about the ingredients piece-- I think this is the direction I would think in too. That said, I'd consider exact search a stretch goal.

app/services/foods_service.rb
```diff
@@ -0,0 +1,17 @@
+class FoodsService
+  def self.conn
+    Faraday.new("https://api.nal.usda.gov/fdc/v1/") do |faraday|
+      faraday.params['api_key'] = ENV['food_api']
+    end
+  end
+
+  def self.search_foods(search)
+    response = conn.get("foods/search?query=#{search}&pageSize=200")
+    JSON.parse(response.body,symbolize_names: true)
```
jamisonordway, line 10: Overall, this class is really well encapsulated and DRY. I do think you could move line 10 to the `conn` method as well.

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,10 @@
+class FoodsFacade
+  def self.food_search(search)
+    food = FoodsService.search_foods(search)
+    FoodInfo.new(food)
+  end
+
+  def self.ten_results(search)
+    # require "pry"; binding.pry
+  end
```
jamisonordway, line 9: The endpoint you're consuming in FoodsService should return everything we need for this story. I image that something like this could return both data points with one call:
```
def self.food_search(search)
  data = FoodsService.search_foods(search)
  total_hits = data[:totalHits]
  foods = data[:results].first(10).map do |food_data|
    FoodInfo.new(food_data)
  end
  [total_hits, foods]
end
```

PR #111, opened 2022-05-27
James Harkins

James-Harkins:


app/facades/foods_facade.rb
```diff
@@ -0,0 +1,19 @@
...
+    until page_counter > 50
+      results << UsdaService.search(ingredient, page_counter)[:foods].find_all do |result|
+        result[:ingredients].include?(ingredient) if result[:ingredients]
+      end
+      page_counter += 1
+    end
+    actual_results = results.find_all do |result|
+      !result.empty?
+    end
+    actual_results.map do |actual_result|
```
jamisonordway, line 14: The facade could be responsible for limiting the number of Food POROs created and returned:
```
actual_results.first(10).map do |actual_result|
  Food.new(actual_result)
end
```

app/services/usda_service.rb
```diff
@@ -0,0 +1,16 @@
...
+    Faraday.new(url: "https://api.nal.usda.gov") do |f|
+      f.headers["api_key"] = ENV["usda_api_key"]
+      f.adapter Faraday.default_adapter
+    end
+  end
+
+  def self.search(query, page)
+    response = conn.get("/fdc/v1/foods/search") do |c|
+      c.options.params_encoder = Faraday::FlatParamsEncoder
+      c.params = {api_key: ENV["usda_api_key"], query: query, pageNumber: page, pageSize: 200}
```
jamisonordway, line 12: There shouldn't be a need to repeat the API key parameter here, since it's already set up as part of every request by the `conn` method.

PR #113, opened 2022-05-27
Sherman Au

ShermanA-13:


app/facades/food_facade.rb
```diff
@@ -0,0 +1,15 @@
...
+      data = FoodService.find_foods(food)
+      data[:foods].map do |food_data|
+        Food.new(food_data)
+      end.take(10)
+    end
+
+    def result_count(food)
+      data = FoodService.find_foods(food)
+      data[:totalHits]
+    end
```
jamisonordway, line 13: - this is creating a lot of Food PORO instances that will never be used/returned to the controller.
- this is hitting the same API endpoint twice. Ideally, we would get both data points with a single call.
```
def food_search(food)
  total_hits = data[:totalHits]
  foods = data[:foods].first(10).map do |food_data|
    Food.new(food_data)
  end
  [total_hits, foods]
end
```

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,6 @@
+class FoodsController < ApplicationController
+  def index
+    @foods = FoodFacade.food_search(params[:q])
+    @result_count = FoodFacade.result_count(params[:q])
```
jamisonordway, line 4: combined with the code snippet I commented on FoodFacade, this could become
```
@result_count, @foods = FoodFacade.food_search(params[:q])
```

PR #115, opened 2022-05-27
John Hennerich turn in 

jhennerich:
I used dream driven development to get the api calls working.  I spent a lot of time getting the api search to work but have a working welcome view page.


app/poros/food.rb
```diff
@@ -0,0 +1,13 @@
+class Food
+  attr_reader :upc,
+              :description,
+              :company,
+              :ingredients
+
+  def initialize(attr)
+    @upc = attr[:gtinUpc] ? attr[:gtinUpc] : 'No response'
```
juliet-e, line 8: Nice edge case thinking here

app/controllers/welcome_controller.rb
```diff
@@ -1,4 +1,11 @@
 class WelcomeController < ApplicationController
   def index
+    results = FoodService.search_for_foods(params[:q])
```
juliet-e, line 3: It seems like you started to pull this logic into a facade but maybe you didn't get to fully refactoring.  Generally t his is exactly what the facade can handle.

juliet-e: Keep in mind as well that we do want to keep the original form on the welcome page, so invoking the food facade would make more sense from a foods controller. With this solution, we don't have a chance to enter query params or fill out the form at all

app/services/food_service.rb
```diff
@@ -0,0 +1,16 @@
+class FoodService
+  def self.conn
```
juliet-e, line 2: Nicely organized service!

app/views/welcome/index.html.erb
```diff
@@ -13,3 +13,12 @@
     </ul>
   </div>
 </nav>
+
+<h1><%= @total_hits %> total results</h1>
+<% @foods[0..9].each do |food| %>
```
juliet-e, line 18: Is there a way to limit the results to 10 without  using ruby logic?

PR #117, opened 2022-05-27
Scott Sullivan Mod3 MidMod IC

ScottSullivanltd:
I did not finish completely but feel pretty good about what I was able to do and my understanding of what was expected.


app/models/food.rb
```diff
@@ -0,0 +1,2 @@
+class Food < Applicationrecord
```
juliet-e, line 1: Is the Food model necessary?

db/migrate/20220527165852_create_foods.rb
```diff
@@ -0,0 +1,12 @@
+class CreateFoods < ActiveRecord::Migration[5.2]
```
juliet-e, line 1: Similar to the comment above re: models -- do we need to store food data at all?

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,19 @@
+require "rails_helper"
+
+RSpec.describe "Food Index Page" do
+  it "displays list of attributes for food search results" do
+    food1 = Food.create(gtinupc: 124, description: "yummy food", brand_owner: "General Mills", ingredients: "hot dog, bun, mustard")
+    food2 = Food.create(gtinupc: 144, description: "yucky food", brand_owner: "General Mills", ingredients: "sweet potatoes")
```
juliet-e, line 6: Rather than creating food in the database, what we're looking for here is you to just retrieve data from the API call. So instead of creating food for test data, you could use a tool like VCR to mock this API call

spec/services/food_service_spec.rb
```diff
@@ -0,0 +1,10 @@
+require "rails_helper"
+
+RSpec.describe FoodService do
+  describe "API repo endpoint" do
+    it "gets food data from  endpoint", :vcr do
```
juliet-e, line 5: Great use of vcr here!

app/services/food_service.rb
```diff
@@ -0,0 +1,8 @@
+class FoodService < BaseService
+  def self.get_food_data
+    response = conn("https://api.nal.usda.gov").get("/fdc/v1/foods/search?api_key=food_api_key&query=")
```
juliet-e, line 3: where is the food_api_key stored? This looks like it's hard-coded rather than pulling from an ENV variable like figaro does. Try to use figaro instead and see if you can get it working.

app/controllers/food_controller.rb
```diff
@@ -0,0 +1,5 @@
+class FoodController < ApplicationController
```
juliet-e, line 1: In order for routing to work, you'll have to make this FoodsController with an s because your resource in routes is titled `foods`

PR #118, opened 2022-05-27
Drew Proebstel

DrewProebstel:


app/facades/foods_facade.rb
```diff
@@ -0,0 +1,10 @@
+class FoodsFacade
+
+  def self.search_by_food(food)
+  	foods_data = FoodsService.foods_by_ingredient(food)[:foods][0..9]
```
juliet-e: Can you find another way to limit the size to 10 without needing to use this ruby logic? (hint: API docs!)

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,6 @@
+class FoodsController < ApplicationController
+  def index
+    @foods = FoodsFacade.search_by_food(params[:q])
+    @hits = FoodsService.foods_by_ingredient(params[:q])[:totalHits]
```
juliet-e: There is a way this could be refactored to only have to make 1 API call. It is also more conventional to have the facade retrieve data from the service (like you did on line 3)

juliet-e: *1 API call for both the 10 foods and the total hits

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  	foods = foods_data[:foods]
+    foods.map do |food_data|
+  	  Food.new(food_data)
+  	end
+  end
+
+  def self.total_hits(food)
+    FoodsService.foods_by_ingredient(food)[:totalHits]
+  end
```
jamisonordway, line 14: Since both of these methods are eventually calling the same endpoint (via FoodService), I would refactor to something like:
```
def self.search_by_food(food)
  foods_data = FoodsService.foods_by_ingredient(food)
  total_hits = foods_data[:totalHits]
  foods = foods_data[:foods].first(10).map do |food_data| do
    Food.new(food_data)
  end
 [total_hits, foods]
end
```

spec/poros/food_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    Food.new({
+      gtinUpc: "123456",
+      description: 'keeps the doctor away',
+      brandOwner: 'earth',
+      ingredients: 'seed, sunlight, water'
+      })
+
+  end
+
+  it 'has attributes', :vcr do
```
juliet-e, line 14: You shouldn't need VCR with a PORO test as its logic doesn't require an external API call. You are passing in the hash and testing that it is creating a ruby object with the right attributes -- that's all you need! No mocking necessary

PR #121, opened 2022-06-01
mid mod review

juliet-e:


app/facade/foods_facade.rb
```diff
@@ -0,0 +1,13 @@
+class FoodsFacade
+  def self.create_food(query)
+    json = FoodsService.get_food_data(query)
+    json[:foods].map do |data|
+      Food.new(data)
+    end.take(10)
```
juliet-e, line 6: You can also limit it to 10 foods by adding the pageSize parameter to the API call. That would remove the need to add this ruby logic here

app/facade/foods_facade.rb
```diff
@@ -0,0 +1,13 @@
...
+  end
+
+  def self.returned_items_number(query)
+    json = FoodsService.get_food_data(query)
+    json[:totalHits]
```
juliet-e, line 11: Same deal as with parks finder. You can add this to the method above and use the same API call to extract the total. Then, just return the foods and the total as an array so the controller can receive both:
`[foods, total] = FoodsFacade.create_food(params[:query])`

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,12 @@
+require "rails_helper"
+
+RSpec.describe "Food Index Page", type: :feature do
+  it "displays list of attributes for food search results", :vcr do
+    food1 = Food.new(gtinupc: 124, description: "yummy food", brand_owner: "General Mills", ingredients: "hot dog, bun, mustard")
```
juliet-e, line 5: Typically, it's more common to pass in the data as a hash as you would expect it to come in from the API data, so all those same fields wrapped in a  { } to indicate that you're passing in a hash that's been parsed from the JSON response

spec/poros/food_spec.rb
```diff
@@ -0,0 +1,11 @@
+require "rails_helper"
+
+RSpec.describe Food do
+  it "exists and has attributes" do
+    data = {description: "Food Test Description"}
```
juliet-e, line 5: Make sure to include all the data you need in the test data, not just the description field

juliet-e (commented):
Great work! I see a lot of improvement and some nice structural changes. I especially like the foundation you've laid with BaseService. I added a couple notes in both repos around refactoring to just use 1 API call. This was a tricky thing that is not strictly necessary to pass the MidMod but is a good opportunity for refactoring for cleaner, more performant code. Tests look a lot better, too. I think your next focus could be on adding more test cases for sad path and edge cases, and you'll get practice with this with consultancy. Great work, overall, Scott!

PR #127, opened 2022-09-09
Alex Boyd - Mid Mod Assessment

alepbloyd:
Completes user story.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+  "workbench.colorCustomizations": {
+    "activityBar.background": "#4A194C",
+    "titleBar.activeBackground": "#68236B",
+    "titleBar.activeForeground": "#FDF8FD"
+  }
+}
--- a/app/controllers/foods_controller.rb
+++ b/app/controllers/foods_controller.rb
@@ -1,0 +1,7 @@
+class FoodsController < ApplicationController
+
+  def index
+    
+  end
+
+end
--- a/app/controllers/foods_controller.rb
+++ b/app/controllers/foods_controller.rb
@@ -1,10 +1,14 @@
 class FoodsController < ApplicationController
 
   def index
-    search_results = FoodFacade.ingredient_search(params[:q])
-    
-    @foods = search_results[0].first(10)
-    @total_result_count = search_results.last
+    if params[:q].empty?
+      redirect_to "/", notice: "Search field cannot be blank!"
+    else
+      search_results = FoodFacade.ingredient_search(params[:q])
+
+      @foods = search_results[0].first(10)
+      @total_result_count = search_results.last
+    end
   end
 
 end
--- a/app/controllers/foods_controller.rb
+++ b/app/controllers/foods_controller.rb
@@ -1,7 +1,10 @@
 class FoodsController < ApplicationController
 
   def index
+    search_results = FoodFacade.ingredient_search(params[:q])
     
+    @foods = search_results[0].first(10)
+    @total_result_count = search_results.last
   end
 
 end
--- a/app/controllers/foods_controller.rb
+++ b/app/controllers/foods_controller.rb
@@ -6,7 +6,8 @@ class FoodsController < ApplicationController
     else
       search_results = FoodFacade.ingredient_search(params[:q])
 
-      @foods = search_results[0].first(10)
+      @foods = search_results.first
+      
       @total_result_count = search_results.last
     end
   end
--- a/app/facades/food_facade.rb
+++ b/app/facades/food_facade.rb
@@ -1,0 +1,13 @@
+class FoodFacade
+
+  def self.ingredient_search(search_string)
+    data = FoodService.foods_by_ingredient(search_string)
+
+    food_results = data[:foods]
+
+    @foods = food_results.map do |food|
+      Food.new(food)
+    end
+  end
+
+end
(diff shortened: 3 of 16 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
```
juliet-e, line 1: Just FYI for the future -- since these are your own personal preferences, it's best practice to add this file to your .gitignore

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,15 @@
+class FoodsController < ApplicationController
+
+  def index
+    if params[:q].empty?
```
juliet-e, line 4: I appreciate your inclusion of error handling!

app/facades/food_facade.rb
```diff
@@ -0,0 +1,15 @@
+class FoodFacade
+
+  def self.ingredient_search(search_string)
+    data = FoodService.foods_by_ingredient(search_string)
+
+    food_results = data[:foods]
+
+    @foods = food_results.map do |food|
```
juliet-e, line 8: Because this is a facade with class methods, you don't need to name these with `@` as they aren't instance variables that need to be accessible elsewhere

app/facades/food_facade.rb
```diff
@@ -0,0 +1,15 @@
...
+      Food.new(food)
+    end
+
+    [@foods.first(10),data[:totalHits]]
```
juliet-e, line 12: This is a great way to handle the case in which you need both of these objects! 

We wouldn't expect anyone to use this in an assessment but just to give you another option for this case: you could use [OpenStruct](https://ruby-doc.org/stdlib-1.9.3/libdoc/ostruct/rdoc/OpenStruct.html) to return both these values. This is used a lot in the industry for this type of case when you could use a PORO but don't necessary need to because you're only creating it as a way to hold these items, not as a domain object to store state like a Food PORO. Instead of creating a whole class, OpenStruct allows you to create a one-time hash-like object where you can store attributes with keys. The benefit of this over an array is that you can then reference, in your controller, `results.food_collection` and `results.total_hits` rather than using `results.first` and `.last`, which can be tricky in case you mess up what order those results come in. 

Again, using an array is a GREAT solution but this is just another option for the future!

app/services/food_service.rb
```diff
@@ -0,0 +1,13 @@
+class FoodService
+
+  def self.conn
+    Faraday.new(url: "https://api.nal.usda.gov/fdc/v1/")
```
juliet-e, line 4: Usually the Faraday connection just contains the *base url*, so all the path components after `.gov` can live in the path on line 8

app/services/food_service.rb
```diff
@@ -0,0 +1,13 @@
...
+  end
+
+  def self.foods_by_ingredient(search_string)
+    response = conn.get("foods/search?query=#{search_string}&api_key=#{ENV['usda_api']}")
```
juliet-e, line 8: you could use the pageSize parameter here to limit foods to 10 so that you don't have to do it in Ruby in your facade

spec/facades/food_facade_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'rails_helper'
+
+RSpec.describe 'food facade' do
+  it 'returns an array of foods given a search string' do
+    results = FoodFacade.ingredient_search("sweet potatoes")
+
+    expect(results).to be_an(Array)
```
juliet-e, line 7: Could have a few more assertions about the contents of the array to make sure you're returning those 2 items in the right order (total hits and the foods)

You could also assert that the foods array only has 10 items

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    fill_in "q", with: "sweet potatoes"
+
+    click_on "Search"
+
+    expect(current_path).to eq("/foods")
+
+    expect(page).to have_content("Total Results: 49676")
+
+    expect(page).to have_selector(".food-item", count:10)
```
juliet-e, line 16: Nice assertion

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  it 'displays the total number of search results for a given term' do
+    visit '/'
+
+    fill_in "q", with: "sweet potatoes"
+
+    click_on "Search"
+
+    expect(current_path).to eq("/foods")
+
+    expect(page).to have_content("Total Results: 49676")
```
juliet-e, line 14: For an assessment this is totally fine. If this were a project I would warn against using this exact value in the assertion unless you're stubbing the API call with Webmock and using a static fixture. Otherwise, if the API changes, this number could change and break your test

spec/poros/food_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'rails_helper'
+
+RSpec.describe Food do
+  it 'exists' do 
+    data = {
+      gtinUpc: "123456789",
```
juliet-e, line 6: You win the award for the best test data

juliet-e:
Excellent submission with solid adherence to MVC, SRP and the design patterns we've learned so far. Great work leveraging Ruby's multiple value return!

app/facades/food_facade.rb
alepbloyd: Oh very cool, I'll definitely check out OpenStruct!

PR #128, opened 2022-09-09
Eli Sachs

easachs:


spec/models/food_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'rails_helper'
+
+RSpec.describe Food do
```
dcoleman21, line 3: So we would want this to be a PORO test vs a Model test. Very similar to how you have it set up in here

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    within '#food-10' do
+      expect(page).to have_content('10. 865648000009')
+      expect(page).to have_content('Description: Sweet potatoes')
+      expect(page).to have_content('Brand Owner: Topashaw farms')
+      expect(page).to have_content('Ingredients: No ingredients found.')
+    end
+  end
+
+  it 'error with no search', vcr: 'error' do
```
dcoleman21, line 26: Nice job sad path testing here.

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+RSpec.describe 'Foods Search' do
+  it 'can search for foods', vcr: 'sweet_potatoes' do
+    visit root_path
+
+    fill_in :q, with: 'sweet potatoes'
+    click_button 'Search'
+
+    expect(current_path).to eq(foods_path)
+    within '#food-1' do
```
dcoleman21, line 11: Nice job using within blocks

spec/facades/foods_facade_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'rails_helper'
+
+RSpec.describe 'FoodsFacade', vcr: 'sweet_potatoes' do
+  it 'returns instances of food with given ingredient' do
```
dcoleman21, line 4: These are great tests. I would think about "Is there a way that we could get all the info we need - totalHits and the attributes we want with a single call?

app/services/foods_service.rb
```diff
@@ -0,0 +1,18 @@
+class FoodsService
+  def self.food_with(ingredient)
+    response = conn.get("/fdc/v1/foods/search?query=#{ingredient}&pageSize=10")
```
dcoleman21, line 3: nice job adding `pageSize=10` here so that you are limiting it to 10 results at the call

app/poros/food.rb
```diff
@@ -0,0 +1,13 @@
+class Food
+  attr_reader :code, :description, :brand, :ingredients
+  def initialize(data)
+    @code = nil_check(data[:gtinUpc]) ? 'No code found.' : data[:gtinUpc]
+    @description = nil_check(data[:description]) ? 'No description found.' : data[:description]
+    @brand = nil_check(data[:brandOwner]) ? 'No brand found.' : data[:brandOwner]
+    @ingredients = nil_check(data[:ingredients]) ? 'No ingredients found.' : data[:ingredients]
+  end
+
+  def nil_check(attribute)
```
dcoleman21, line 10: Nice helper method here to check for nil

app/poros/food.rb
```diff
@@ -0,0 +1,13 @@
+class Food
+  attr_reader :code, :description, :brand, :ingredients
+  def initialize(data)
+    @code = nil_check(data[:gtinUpc]) ? 'No code found.' : data[:gtinUpc]
```
dcoleman21, line 4: Great job using a ternary expression to check for nil

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,11 @@
+class FoodsFacade
+  def self.food_with(ingredient)
```
dcoleman21, line 2: So here is where we can make one method that gets us all that we want/need. We can set line 3 to a variable 
```
food_results = FoodService.food_with(ingredient)
```
Then we will iterate over our new variable and set this to another variable
```
food_attrs = food_results[:foods].map do |food_data|
Food.new(food_data)
```
Notice I took out the `[0..9]` range because you already limited your results to 10 with `pageSize=10` in your Service.

Finally, the last line that we want returned, outside of the block is to create a nested array that has both totalHits and all the attributes that we want.
```
total_hits, foods = [food_results[:totalHits], food_attrs]
```
This will give us the ability to have access to both things in our Controller. 

So all put together it could look like this:
```
      Food.new(food_result)
    end
end
```

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,7 @@
+class FoodsController < ApplicationController
+  def index
```
dcoleman21, line 2: See my comment in the Facade. How can we get all that we need by making a single call vs calling two separate methods?

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,7 @@
...
+    @query = params[:q]
+    @foods = FoodsFacade.food_with(params[:q])
+    @total = FoodsFacade.total_with(params[:q])
```
dcoleman21, line 5: If you followed the comments in the Facade, your controller could look like:
```
results = FoodFacade.food_with(params[:q])
@total_hits = results[0]
@foods = results[1]
```

dcoleman21 (commented):
Overall nice job Eli. I left some comments for you to look over. Let me know if you have any questions.

PR #129, opened 2022-09-09
Wesley Miranda

wes30303:


app/controllers/application_controller.rb
```diff
@@ -1,3 +1,9 @@
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
+  before_action :api_request
+
+  def api_request
```
juliet-e, line 5: We typically don't use before_actions for the main logic in controllers but rather for operations before getting to the main logic (i.e. making sure a user is logged in before making an api call or querying the database)

app/controllers/application_controller.rb
```diff
@@ -1,3 +1,9 @@
...
+    @search = TmdbSearch.new
```
juliet-e, line 6: Facades usually have class methods so they shoudln't require instantiation.

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,5 @@
+class FoodsController < ApplicationController
+  def index
+    @foods = @search.keyword(params[:name])
```
juliet-e, line 3: Instead of relying on the callback_action, consider just calling the facade here. Also make sure to name things properly and not based on your old project! Can be named `FoodFacade` instead of TmdbSearch

app/facads/tmdb_search.rb
```diff
@@ -0,0 +1,11 @@
+class TmdbSearch
+
+  def service
+    TmdbService.new
```
juliet-e, line 4: Services should have class methods so you don't need to instantiate them with `.new`

app/facads/tmdb_search.rb
```diff
@@ -0,0 +1,11 @@
...
+  end
+
+  def keyword(name)
+    @_search = JSON.parse(service.get_url.get("/fdc/v1/foods/search?query#{name}").body, symbolize_names: true)
```
juliet-e, line 8: Does the facade need to know the API search path or can all the API logic live in the service?

app/services/tmdb_service.rb
```diff
@@ -0,0 +1,8 @@
+class TmdbService
+
+ def get_url
```
juliet-e, line 3: It's conventional for the service to create the connection *and* have the method that will make a get request. All the API details should stay together in one class

app/poro/food.rb
```diff
@@ -0,0 +1,7 @@
+require 'pry'
+
+class Food
+  def initialize(attributes)
```
juliet-e, line 4: I would expect to see assignment of attributes here that are used in the view

PR #132, opened 2022-09-09
Mary Turpin

MaryT573:
Consumed the api, but stalled out on routing the search from the prebuilt form to the food controller. Gave up there


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,12 @@
+class FoodsController < ApplicationController
+    def index
+        @food = FoodFacade.search_list_by_food(params[:search])
+
+    end
+
+    private
+
+    def food_params
```
juliet-e, line 9: Strong params are needed when we are gathering user input from something like a form. In this case, are strong params needed for name, brand, code, ingredients?

app/models/food.rb
```diff
@@ -0,0 +1,3 @@
+class Food < ApplicationRecord
```
juliet-e, line 1: Keep in mind that Models persist in the database while POROs do not. When we are retrieving data from an API and displaying it, we usually don't need to persist that data in the database, so we just need a PORO, not a model

app/poros/food.rb
```diff
@@ -0,0 +1,10 @@
+class Food
```
juliet-e, line 1: Nice PORO structure!

app/services/food_service.rb
```diff
@@ -0,0 +1,15 @@
...
+ 
+    def self.search_food(food)
+        response = conn.get("search?query=#{food}")
+        JSON.parse(response.body, symbolize_names: true) 
+      end
+
+    private  
+
+    def self.conn
+        Faraday.new(url: "https://api.nal.usda.gov/fdc/v1/foods/") do |faraday|
```
juliet-e, line 11: Typically, the Faraday connection would set up the connection with *just* the base URL, so everything up until .gov. The `/fdc/v1/foods/search` part would be in the path on line 4. In this setup, you have part of the path here and part on line 4, which is just a bit less conventional and can cause problems

config/routes.rb
```diff
@@ -1,3 +1,4 @@
 Rails.application.routes.draw do
   root "welcome#index"
+  get "/food", to: "foods#index"
```
juliet-e, line 3: if you do `/foods` instead, that would allow the form to work without the change you made from `form_tag` to `form_with`. the `:foods` in the form_tag line denotes the path that it would direct things too (/foods)

spec/poros/food_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'rails_helper'
+
+RSpec.describe Food do
+    it 'reads data into variables' do
```
juliet-e, line 4: Nice PORO test! If this were a project it would probably make sense to put the data stub in a fixture since the ingredients line is so long, but for the purpose of this assessment it's totally fine

spec/services/food_service_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'rails_helper'
+
+RSpec.describe FoodService do
+    it 'gets data from api' do
+        foods = FoodService.search_food('cheese') 
+ 
+        expect(foods).to be_a(Hash)
```
juliet-e, line 7: We'd expect to see more assertions here around the presence of certain keys that will then be used in the PORO

PR #134, opened 2022-09-09
Drew MacNicholas

dmacnicholas:


app/facades/food_facade.rb
```diff
@@ -0,0 +1,9 @@
+class FoodFacade
+
+  def self.find_food(food)
+    food_finder = FoodService.find_food(food)
+    food_finder[:foods].map do |data|
```
juliet-e, line 5: You can also retrieve the totalHits here (consider that in Ruby you can return more than one object from a Ruby method)

juliet-e: This allows you to not have to save it to each individual PORO but return the data from this method like this:
`[foods, food_finder[:totalHits]]`

app/services/food_service.rb
```diff
@@ -0,0 +1,12 @@
+class FoodService
+  def self.conn
+    Faraday.new(url:'https://api.nal.usda.gov/fdc/v1/') do |faraday|
```
juliet-e, line 3: Be careful about including part of the path in the faraday connection as that can sometimes cause some weird bugs. It's conventional to just add the base url (i.e. everything up to .gov) but I'd add /fdc/ to the path on line 10

spec/facades/food_facade_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'rails_helper'
+
+RSpec.describe FoodFacade do
+      it 'returns Food data', :vcr do
+        food_results = FoodFacade.find_food("sweet potatoes")
+
+        expect(food_results.first.gtin_upc).to eq("8901020020844")
```
juliet-e, line 7: This can be a bit tricky in case the first food changes in the API and VCR re-records -- it might change. For the purpose of the assessment this is totally fine/reasonable but in a project I'd suggest testing exact values in the POROs instead where you can easily stub the data and make assertions about attributes getting populated with the right thing

spec/facades/food_facade_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+        expect(food_results).to be_a(Array)
+        food_results.each do |result|
+          expect(result).to be_a(Food)
```
juliet-e, line 10: You could also add an assertion about having only 10 foods in the array

spec/rails_helper.rb
```diff
@@ -62,3 +62,9 @@
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
 end
+
+VCR.configure do |config|
+  config.cassette_library_dir = 'spec/fixtures/vcr_cassettes'
+  config.hook_into :webmock
+  config.configure_rspec_metadata!
```
juliet-e, line 69: Remember to filter out your API key in the future! But nice work getting VCR up and running for this assessment!

spec/services/food_service_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'rails_helper'
+
+RSpec.describe FoodService do
+  it 'returns food data', :vcr do
```
juliet-e, line 4: Super robust test -- very nice!

PR #136, opened 2022-09-09
Gauri Joshi

gaurijo:


app/services/food_service.rb
```diff
@@ -0,0 +1,12 @@
+class FoodService
+  def self.conn 
+    Faraday.new("https://api.nal.usda.gov/fdc/v1/foods") do |faraday|
+      faraday.params['api_key'] = ENV['food_api_key']
+    end
+  end
+
+  def self.get_foods(ingredient)
+    response = conn.get('list', {query: ingredient})
```
juliet-e: We were actually expecting you to hit the `/search` endpoint for this user story. While list is okay for the food items (more or less the same data) it doesn't provide the total number of hits we were looking for. We know it's hard to spend a lot of time reading the docs during an assessment though, so in the future when facing this question of which endpoint to use, when in a non-assessment format, it would be worth doing a bit more research to find the endpoint that returns all the data you need

app/services/food_service.rb
```diff
@@ -0,0 +1,12 @@
+class FoodService
+  def self.conn 
+    Faraday.new("https://api.nal.usda.gov/fdc/v1/foods") do |faraday|
```
juliet-e: Typically the Faraday connection will just include the base URL, so everything after `.gov` can be moved to the path on line 9

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,6 @@
+<% @foods.each do |food|%>
```
juliet-e, line 1: Just  missing the total hits here

spec/facade/food_facade_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'rails_helper'
+
+RSpec.describe FoodFacade do 
+  it "gets data for food by ingredient", :vcr do 
+    foods = FoodFacade.create_foods('sweet potatoes')
+
+    expect(foods[0]).to be_a(Food)
```
juliet-e: You can also make assertions about how many Food items are in the array (10). that the top level data structure is an array, etc. 

We would also want you to add an assertion about the presence of the total hits in your returned data

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'rails_helper'
+
+RSpec.describe "Food Index" do 
+  it "has the top 10 food items that contain this ingredient", :vcr do 
+    visit root_path 
+    # fill_in "Ingredient Search", with: 'sweet potatoes' #getting capybara disabled error
+    click_button 'Search'
+
+    expect(current_path).to eq("/foods")
+    expect(page).to have_content('Description: ')
```
juliet-e, line 10: Nice work making these assertions stable enough to not break if that data from VCR changes!

spec/poros/food_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'rails_helper'
+
+RSpec.describe Food do 
+  it "exists and has attributes" do 
+    response = JSON.parse(File.read('spec/fixtures/food.json'), symbolize_names: true)
```
juliet-e, line 5: Great work using a fixture! Makes for a very clean PORO test.

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'rails_helper'
+
+RSpec.describe "Food Index" do 
+  it "has the top 10 food items that contain this ingredient", :vcr do 
+    visit root_path 
+    # fill_in "Ingredient Search", with: 'sweet potatoes' #getting capybara disabled error
```
juliet-e: Try this syntax: `fill_in :q, with: "sweet potatoes"` --> the form field is Q rather than Ingredient search

juliet-e: If that doesn't work, I'd love to see the error you're getting!

spec/services/food_service_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'rails_helper'
+
+RSpec.describe FoodService do 
+  it "gets data for food ingredient", :vcr do 
+    foods = FoodService.get_foods("sweet potatoes")
+    # require 'pry'; binding.pry 
+    expect(foods).to be_a(Array)
```
juliet-e: I'd expect the assertions to be a bit more extensive, i.e. testing for the presence of the keys you would need in the PORO, testing the presence of totalHits, testing the presence of the `foods` array, etc

PR #138, opened 2022-09-09
Bryan Shears

b-shears:


app/facades/food_facade.rb
```diff
@@ -0,0 +1,8 @@
+class FoodFacade 
+    def self.food_finder(food)
+        food = FoodService.search_food(food)
+        food[:foods].map do |food_data|
+            Food.new(food_data, food[:totalHits])
```
juliet-e, line 5: Clever solution to grabbing the totalHits, but there's actually an easier way!

Keep in mind that Ruby allows you to return more than one object from a method, so you could consider returning both your Foods array and food[:totalHits]

I appreciate that you did try to do this in just one API call though -- nice work trying to minimize your external calls to Food API!

app/services/food_service.rb
```diff
@@ -0,0 +1,12 @@
+class FoodService 
+    def self.conn 
+        Faraday.new("https://api.nal.usda.gov/fdc/v1/") do |faraday| 
+            faraday.params['api_key'] = ENV['food_api_key']
+        end 
+    end 
+
+    def self.search_food(food)
+        response = conn.get("foods/search?query=#{food}&pageSize=10&pageNumber=1")
```
juliet-e, line 9: Nice work taking advantage of the pageSize parameter so that you don't have to limit the responses on your end!

spec/facades/food_facade_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'rails_helper'
+
+RSpec.describe FoodFacade, :vcr do 
+
+    it 'returns food data of the searched food' do 
+        search = FoodFacade.food_finder("sweet potatoes") 
+
+        expect(search).to be_a(Array)
```
juliet-e, line 8: I'd expect an assertion here about what type of objects are in the array (i.e. Food objects)

spec/features/foods/foods_index_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'rails_helper'
+
+RSpec.describe 'foods index page', :vcr do
```
juliet-e, line 3: Great robust assertions and nice work using VCR! If this were a project I would be worried about this test being delicate (i.e. if the API adds a food, and VCR re-records, then total hits might go up to 49677 and break this test) but for an assessment this is a great level of detail

spec/poros/food_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+RSpec.describe Food do 
+    xit 'exists' do 
+        food = Food.new(total_hits: 1, description: "liter of cola", code: "123", brand_owner: "Coca-Cola", ingredients: "crushed dreams")
+        expect(food).to be_an_instance_of(Food)
```
juliet-e, line 6: Saw your note about brain fart with the # of arguments here -- makes sense! We would just expect to see assertions about the exact attributes in the food object

spec/rails_helper.rb
```diff
@@ -62,3 +62,10 @@
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
 end
+
+VCR.configure do |config|
+  config.cassette_library_dir = 'spec/fixtures/vcr_cassettes'
+  config.hook_into :webmock
+  config.configure_rspec_metadata!
```
juliet-e, line 69: Remember to filter out your API key in the future!

spec/services/food_service_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'rails_helper'
+
+RSpec.describe FoodService, :vcr do
```
juliet-e, line 3: Awesome test

PR #139, opened 2022-09-09
Thomas Turner

EagleEye5085:
my tests need to be more robust. most of my time was spent on the api setup. i made basic tests to get me the data i needed.


config/routes.rb
```diff
@@ -1,3 +1,4 @@
 Rails.application.routes.draw do
   root "welcome#index"
+  resources :foods
```
dcoleman21, line 3: I would add ` only: [:index]` since that is the only action we have

spec/services/usda_service_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'rails_helper'
+
+RSpec.describe 'usda service' do
```
dcoleman21, line 3: Per your comment, yes, more robust testing would be good. And in the Service and Facade we can test for data types. Also, we will want to use vcr or webmock to stub out the calls

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    fill_in :q, with: 'sweet potatoes'
+
+    click_button 'Search'
+
+    expect(current_path).to eq(foods_path)
+
+    expect(page).to have_content('Total hits: 49676')
+
+    expect(page).to have_content('Item GTIN/UPC code: 8901020020844')
```
dcoleman21, line 16: We are wanting to test that we have a list of 10 foods and all of their attributes not just the upc

spec/facades/usda_facade_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'rails_helper'
+
+RSpec.describe 'usda facade' do
```
dcoleman21, line 3: Same here. More robust, data types and use vcr or webmock

app/services/usda_service.rb
```diff
@@ -0,0 +1,19 @@
+class UsdaService
+
+   def self.get_url
+      Faraday.new("https://api.nal.usda.gov") do |faraday|
+        faraday.params["api_key"] = ENV['usda_key']
+      end
+    end
+
+    def self.name_search(name)
+      @_name_search = JSON.parse(get_url.get("/fdc/v1/foods/search?query=#{name}").body, symbolize_names: true)
```
dcoleman21, line 10: I think you could make a single call here and just add params for pageSize = 10. This would give you all the info you need in a single method.

app/facades/usda_facade.rb
```diff
@@ -0,0 +1,13 @@
+class UsdaFacade
+
+  def self.name_search_items(name)
+    UsdaService.name_search_items(name).map do |food|
+      Food.new(food)
+    end
+  end
+
```
dcoleman21, line 8: Same thing here. If you add additional params (see note in your service) I think you could have a single method that gives you all that you want.

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,6 @@
+class FoodsController < ApplicationController
+  def index
```
dcoleman21, line 2: This is clean. How can we slim it up even more? :)

dcoleman21 (commented):
Overall, nice job. I have left comments for you to review. Let me know if you have any questions.

PR #140, opened 2022-09-09
Parker Thomson

pdthomson:


app/facades/foods_facade.rb
```diff
@@ -0,0 +1,16 @@
+class FoodsFacade
+
+  class << self
+
+    def get_foods(food)
+      json = FoodsService.find_food(food)[:foods]
+
+      json[0..9].map {|food| Food.new(food)}
```
juliet-e, line 8: You can also limit the # of responses back from the API using the `pageSize` query parameter it provides. Another option to avoid having to limit it with Ruby here!

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,16 @@
...
+    end
+
+    def count_foods(food)
+      json = FoodsService.find_food(food)[:totalHits]
```
juliet-e, line 12: Can you figure out how to grab total hits with just one API call (i.e. get this data point and the array of foods in one method)

hint: you can return more than one object from Ruby methods. 
Next week, we'll also learn about caching which would make your current setup awesome because you could keep the separate of getting foods and counting foods to separate methods, but you would retrieve totalHits from the cache rather than having to make another API call

app/services/foods_service.rb
```diff
@@ -0,0 +1,20 @@
+class FoodsService
+
+  class << self
+
+    def get_url(url, keyword = nil)
+      conn = Faraday.new(url: 'https://api.nal.usda.gov') do |faraday|
+        faraday.params[:api_key] = ENV['food_api_key']
+        faraday.params[:query] = keyword unless keyword.nil?
```
juliet-e, line 8: Nice nil check here!

app/poros/food.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  attr_reader :upc,
+              :description,
+              :brand_owner,
+              :ingredients
+
+  def initialize(data)
+    binding.pry
+
+    @upc = data[:gtinUpc]
```
juliet-e, line 11: To tighten your logic up even more -- there are some foods where gtinUpc and brandOwner aren't present in their hash, so you could bring this to the next level by adding a nil check to make sure the JSON key is present before accessing its value

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,12 @@
+"Sweet Topatoes"
```
juliet-e, line 1: Can you make this dynamic in case the query isn't Sweet Topatoes every time?

spec/facades/foods_facade_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'rails_helper'
+
+RSpec.describe 'Food Facade' do
+
+  it "returns fods containing a keyword" do
+    foods = FoodsFacade.get_foods('sweet potatoes')
+
+    expect(foods).to be_an(Array)
+    expect(foods).to be_all(Food)
```
juliet-e, line 9: you can also add an assertion here about how many Food objects are in the array (always 10)

spec/facades/foods_facade_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  it "returns the count of foods containing the key word" do
+    count = FoodsFacade.count_foods('sweet potatoes')
+
+    expect(count).to eq(49676)
```
juliet-e, line 15: For an assessment this is totally fine, but generally I would caution against expecting the count to be an exact value because that can easily change over time with this API, breaking your test. You could instead just make sure it's an Integer, or that it's >40,000

spec/poros/food_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+      description: 'sweet potatoes',
+      gtinUpc: '492111402857',
+      brandOwner: 'ARCHER FARMS',
+      ingredients: 'SWEET POTATOES'
+    }
+
+    food_info = Food.new(attributes)
+
+    expect(food_info).to be_an_instance_of(Food)
+    expect(food_info.description).to eq('sweet potatoes')
```
juliet-e, line 16: Nice assertions here -- Here it's totally fine to expect the exact values because we are stubbing the hash we pass in. So if you use fixtures, or use just webmock in other tests, exact value testing is fine for external API data. But, if you use VCR or don't have a mock set up, then that data is likely to change often and might break your tests if you expect exact values. 

But again, for the purposes of an assessment, it's very reasonable to test in this way as there is limited time.

PR #142, opened 2022-09-09
Jared Hardinger

jaredhardinger:


spec/features/search_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    - the food's description
+    - the food's Brand Owner
+    - the food's ingredients" do 
+        visit root_path
+        fill_in 'q', with: 'sweet potatoes'
+        click_button('Search')
+        expect(current_path).to eq(foods_path)
+        
+        expect(page).to have_content("Total Number of Items Returned by Search:")
+        expect(page).to have_content("GTIN UPC:")
```
dcoleman21, line 20: It might be a good idea to put these expect for attributes inside of a `within block`. It could look like:
```
within(first('.foods')) do
      expect(page).to have_css(".gtinUpc")
      expect(page).to have_css(".description")
      expect(page).to have_css(".brandOwner")
      expect(page).to have_css(".ingredients")
end
```
We would also want to test that we should be getting 10 results back.

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,6 @@
+require 'rails_helper'
+
+RSpec.describe 'The Foods Index Page' do
```
dcoleman21, line 3: You may not need this if you are testing for the attributes like you did in the `search_spec`

config/routes.rb
```diff
@@ -1,3 +1,4 @@
 Rails.application.routes.draw do
   root "welcome#index"
+  resources :foods
```
dcoleman21, line 3: I would probably add the `, only: [:index]` since it is the only action we have

app/services/foods_service.rb
```diff
@@ -0,0 +1,18 @@
+class FoodsService
+    def self.get_url(url, keyword = nil)
+        conn = Faraday.new(url: "https://api.nal.usda.gov") do |f|
+            f.headers[:'x-api-key'] = ENV['api_key']
+            f.params[:query] = keyword unless keyword.nil?
+        end
```
dcoleman21, line 6: Nice check for nil value

app/services/foods_service.rb
```diff
@@ -0,0 +1,18 @@
+class FoodsService
+    def self.get_url(url, keyword = nil)
+        conn = Faraday.new(url: "https://api.nal.usda.gov") do |f|
+            f.headers[:'x-api-key'] = ENV['api_key']
+            f.params[:query] = keyword unless keyword.nil?
```
dcoleman21, line 5: you might also add the param for [:totalHits] here as well. So that you don't have to make two separate calls.

app/facades/food_facade.rb
```diff
@@ -0,0 +1,17 @@
+class FoodFacade
+    def self.food_search(keyword)
```
dcoleman21, line 2: Nice job setting up the facade. How could we make one call and get both the keyword search results and a total hits?

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,7 @@
+class FoodsController < ApplicationController
+    def index
+        @keyword = params[:'q']
+        @foods = FoodFacade.food_search(@keyword).first(10)
```
dcoleman21, line 4: If you include params for `:totalHits` and set it to 10 in your service, you won't need to use `.first(10)`
Think about how you could refactor this to make one call to the Facade and get both count and food results.

dcoleman21 (commented):
I didn't see any tests for your Facade, Service, PORO

PR #143, opened 2022-09-09
Jonathan Pope

jonathanmpope:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,12 @@
+class FoodsController < ApplicationController 
+
+    def index  
+        if params[:q] != nil && params[:q] != ''
+            @total_foods = FoodsIndexSearchTotalFacade.service(params[:q]) 
+            @foods = FoodsIndexSearchResultFacade.service(params[:q])  
+        else  
+            flash[:error] = "Please submit a query in the search form"
```
juliet-e, line 8: Nice addition of error handling!

app/facades/foods_index_search_result_facade.rb
```diff
@@ -0,0 +1,10 @@
+class FoodsIndexSearchResultFacade
+
+    def self.service(search)
```
juliet-e, line 3: This is nit, more sharing for context/info: Typically, facade method names are named after the data object they are providing. So it's more often seen that this would be called `food_results` or something like that

app/facades/foods_index_search_total_facade.rb
```diff
@@ -0,0 +1,8 @@
+class FoodsIndexSearchTotalFacade
+
+    def self.service(search) 
+        data = FoodService.food_search(search)
+       
+        @total_foods = data[:totalHits]
```
juliet-e, line 6: Consider trying to minimize the # of API calls you are making. Without caching (which we will learn about in week 5), we're making 2 of the same calls here. But you could refactor this to grab the total hits data in the first call (keep in mind that Ruby allows you to return more than one object from a method)

app/services/food_service.rb
```diff
@@ -0,0 +1,14 @@
+class FoodService
+
+    def self.conn 
+        Faraday.new(url: "https://api.nal.usda.gov") do |faraday|
+            faraday.headers["api_key"] = ENV['usda_key']
+        end
+    end 
+
+    def self.food_search(search)
+        response = conn.get("/fdc/v1/foods/search?api_key=#{ENV['usda_key']}&query=#{search}&pageSize=10&pageNumber=1")
```
juliet-e, line 10: you are hitting the endpont we were expecting you to hit!

spec/features/search_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+        it 'shows the number of returned results' do
+            visit '/'
+            
+            fill_in "q", with: "sweet potatoes"
+            click_button("Search")
+
+            expect(current_path).to eq("/foods")
+
+            expect(page).to have_content("Number of results: 49676")
```
juliet-e, line 22: Typically I would warn against making assertions on exact values given that this data could change and break your test (if you're not mocking or using VCR to mock). For an assessment this is fine, but for exact value assertions, webmock or stubbing what comes back from a service call is a safer option for more stable testing.

PR #144, opened 2022-09-09
Shirley DeCesari

sdecesari:


config/routes.rb
```diff
@@ -1,3 +1,4 @@
 Rails.application.routes.draw do
   root "welcome#index"
+  resources :foods, only: [:index]
```
dcoleman21, line 3: This looks great

app/views/foods/index.html.erb
```diff
@@ -0,0 +1,8 @@
+<h1>Search Results Found: <%= @food_count %></h1>
+
+<% @foods.each do |food| %>
+  <h3>UPC Code: <%= food.upc %></h3>
```
dcoleman21, line 4: Nice job iterating here and getting all of the attributes for each food object

app/services/foods_service.rb
```diff
@@ -0,0 +1,15 @@
+class FoodsService
+  def self.get_list_of_foods(q)
+  end_point = '/fdc/v1/foods/search'
+    response = conn.get(end_point) do |faraday|
+      faraday.params['api_key'] = ENV['food_api_key']
+      faraday.params['query'] = q
```
dcoleman21, line 6: I think you could have also added a param for `pageSize` and set it to 10

app/poros/food.rb
```diff
@@ -0,0 +1,12 @@
+class Food
```
dcoleman21, line 1: this looks good.

app/poros/food.rb
```diff
@@ -0,0 +1,12 @@
+class Food 
+  attr_reader :upc,
+              :description,
+              :brand_owner,
+              :ingredients
+  def initialize(data)
+    @upc = data[:gtinUpc]
```
dcoleman21, line 7: For UPC you could also add a check for if there is nothing returned. 

This is called a ternary and it might look like this:
```
@upc = data[:gtinUpc] ? data[:gtinUpc] : 'No response'
```
Or you could create a helper method that does the same as above like this:
```
def upc_format(data)
   if data[:gtinUpc]
        data[:gtinUpc]
   else
         'No response`
   end
end    
```
If you use to helper method, then in your initialize you would call this:
```
@upc = upc_format(data)      
```

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,12 @@
+class FoodsFacade
+  def self.get_list(q)
```
dcoleman21, line 2: Here we are making two calls to the api, one to get the food list and another to get a count. This is where adding that param for `:totalHits` comes into play.
So here is an example of how you might be able to make one call and get both bits:
```
      Food.new(food_result)
    end
  end
end
```
the `total_hits, foods = [food_results[:totalHits], x]` is creating an array with our keyword search and total hits. Which we can call in the controller. See controller note

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,6 @@
+class FoodsController < ApplicationController
+  def index
```
dcoleman21, line 2: So following the note in your facade, where we created an array that holds both the foods from our keyword search and the total_hits, we could access those via one call like such:
```
results = FoodFacade.get_list(params[:q])
@foods = results[1]
@total_hits = results[0]
```

dcoleman21 (commented):
I have left comments throughout. I don't see your spec/features or any tests for Facades, Services or PORO???

PR #146, opened 2022-09-09
Nicole E. - 2205 M3 Mid Mod

nicecodingtoo:


app/facades/food_facade.rb
```diff
@@ -0,0 +1,13 @@
+class FoodFacade
```
juliet-e, line 1: Typically, facade and services use class methods so that you don't need to instantiate them!

They are sort of utility classes (they do stuff for us) rather than classes that need to hold any state (like a PORO holding state in attributes). That makes it more reasonable for these methods to be class methods rather than instance methods

app/services/food_service.rb
```diff
@@ -0,0 +1,18 @@
+class FoodService
+   def food_search(food)
+      response = conn.get("foods/search?query=#{food}&pageSize=10&pageNumber=1")
+      body = parsed(response)
+   end
+
+   private
+
+   def conn 
+      Faraday.new(url:'https://api.nal.usda.gov/fdc/v1/') do |faraday|
```
juliet-e, line 10: Be careful about including part of the path in the faraday connection as that can sometimes cause some weird bugs. It's conventional to just add the base url (i.e. everything up to .gov) but I'd add `/fdc/v1/` to the path on line 3

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'rails_helper'
+
+RSpec.describe 'food search page' do
+   it 'see a total of the number of items returned by the search' do
+      visit root_path
+      fill_in :q, with: "sweet potatoes"
+      click_on "Search"
+
+      expect(current_path).to eq(foods_path)
+      expect(page).to have_content("Total number of items found: 10")
```
juliet-e, line 10: Instead of listing the total items we're displaying (which will always be 10) we actually wanted you to display the total number of matches from the API, which is stored at the `totalHits` key in the json response. 

Figuring out how to capture that data is a bit tricky so I recommend thinking a bit about how you could grab that data point and make it accessible in the view

spec/facades/foods_facade_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'rails_helper'
+
+RSpec.describe 'Food Facade' do
+   it 'get_food_search' do
+      search = FoodFacade.new
+      search_results = search.get_food_search("sweet potatoes")
+
+      expect(search_results).to be_a(Array)
```
juliet-e, line 8: Could include assertions here about the contents of the array (i.e. all Food items, only 10 of them, etc)

PR #147, opened 2022-09-09
Rue Zheng

ruezheng:


app/facades/food_facade.rb
```diff
@@ -0,0 +1,10 @@
+class FoodFacade
+
+  def self.create_food_search(keyword)
+    response = FoodService.get_food_search(keyword)
+
+    response[:foods][0..9].map do |food_data|
```
juliet-e, line 6: Consider using the API's query parameter of pageSize to limit results to 10 so that you don't have to use ruby here to do that (just a bit cleaner)

app/poros/food.rb
```diff
@@ -0,0 +1,15 @@
+class Food
+  attr_reader :total_hits,
```
juliet-e, line 2: Adding total_hits here is a clever solution to grabbing that data. Keep in mind that you can also return more than one object from a ruby method, so in your facade, you could return `[foods, food_data[:totalHits]` and then use `.first` and `.last` in you controller to distinguish between those 2 objects

app/services/food_service.rb
```diff
@@ -0,0 +1,22 @@
+class FoodService
+
+  def self.conn
+    Faraday.new(url: 'https://api.nal.usda.gov/fdc/') do |faraday|
```
juliet-e, line 4: Be careful about including part of the path in the faraday connection as that can sometimes cause some weird bugs. It's conventional to just add the base url (i.e. everything up to .gov) but I'd add `/fdc/` to the path on line 10

spec/facades/food_facade_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'rails_helper'
+
+RSpec.describe FoodFacade do
```
juliet-e, line 3: Nicely structured facade tests with great assertions

spec/features/foods/search_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'rails_helper'
+
+RSpec.describe 'Foods Search', :vcr do
+  it 'can fill in the search form and see the total number of items returned by the search' do
+    visit '/foods'
```
juliet-e, line 5: If you just visit `/foods` without filling out the form at the root path, you won't have a search query. So this test doesn't pass because the assertion on line 7 will have an empty value for the query (i.e. it won't say sweet potatoes)

spec/poros/food_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'rails_helper'
+
+RSpec.describe Food do
+  it 'exists and has the correct attributes' do
+    # response = JSON.parse(File.read('spec/fixures/food_search_response.json'), symbolize_names: true)
```
juliet-e, line 5: I appreciate that I can tell you considered both a fixture and an abridged attribute stubbed hash :)

spec/services/food_service_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'rails_helper'
+
+RSpec.describe FoodService do
+  describe '::get_food_search', :vcr do
+    it 'gets data for food names that match a keyword(s)' do
+      response = FoodService.get_food_search('sweet potatoes')
+      
+      expect(response[:totalHits]).to be > 30000
```
juliet-e, line 8: I like this assertion because it's not as flaky as testing the exact value, which could change, but it also doesn't just test the data type to have more precision in know the result was more or less what we expected!

PR #148, opened 2022-09-09
Ethan Nguyen

Ethan-t-n:


app/facades/foods_facade.rb
```diff
@@ -0,0 +1,18 @@
+class FoodsFacade
+  def self.service
+    @_service ||= FoodsService.new
```
juliet-e, line 3: Service methods are conventionally class methods too -- no need to instantiate

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  def self.search_by_keyword(word)
+    filtered_results = []
+    
+    service.search_by_keyword(word)[:foods].each do |food|
+      if food[:ingredients].downcase.include?(word)
```
juliet-e, line 10: Nice filter as an extension! An extra challenge would be to find a solution that can do the filter and map the JSON items into POROs in just *one* enumerable rather than iterating through the whole list twice (which can be a time-consuming operation)

app/services/foods_service.rb
```diff
@@ -0,0 +1,11 @@
+class FoodsService
+  def get_url(url)
+    conn = Faraday.new("https://api.nal.usda.gov/fdc/v1/foods/")
```
juliet-e, line 3: Typically the faraday connection can just have the base url (everything up until .gov) and then the path can be in the get call

app/services/foods_service.rb
```diff
@@ -0,0 +1,11 @@
+class FoodsService
+  def get_url(url)
```
juliet-e, line 2: Most conventional to make these all class methods as it is a utility class rather than a class that needs to hold state (like a Food PORO for example)

spec/features/welcome_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'rails_helper'
+
+RSpec.describe "Welcome Page" do
```
juliet-e, line 3: Nice robust test assertions here! If this were a project I would worry about how delicate these tests are given that you're using VCR and the data could change (ie. they could add a food result and that number could go up to 49652) but for an assessment this is a great level of robustness

spec/poros/food_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'rails_helper'
+
+RSpec.describe Food do
+  it "exists", :vcr do
+    food = Food.new(gtin_upc: 123, description: "coffee", brand_owner: "Coffee Co.", ingredients: "coffee")
+
+    expect(food).to be_an_instance_of(Food)
```
juliet-e, line 7: Here I would expect to see assertions about the exact value of the attributes. Because you're stubbing the food data above, you also don't need to be worried about that data changing like it could with VCR so you could really test that each attribute is what it should be

spec/rails_helper.rb
```diff
@@ -62,3 +62,10 @@
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
 end
+
+VCR.configure do |config|
+  config.cassette_library_dir = 'spec/fixtures/vcr_cassettes'
+  config.hook_into :webmock
+  config.configure_rspec_metadata!
+  config.filter_sensitive_data('<api_key>') { ENV['api_key'] }
```
juliet-e, line 70: Great work setting up VCR and filtering the API key!

spec/services/food_service_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'rails_helper'
+
+RSpec.describe FoodsService, :vcr do
+  it "search by keyword" do
+    service = FoodsService.new
+    foods = service.search_by_keyword("sweet potatoes")[:foods]
+
+    expect(foods).to be_an(Array)
+    expect(foods.first).to be_a(Hash)
+    expect(foods.first).to have_key(:ingredients)
```
juliet-e, line 10: It's a good idea to just check that all the necessary keys are there. Consider using `include` to test the presence of all the keys you need to use in the PORO

PR #149, opened 2022-09-09
Nick Jones Submission

nlj77:


spec/services/food_service_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'rails_helper'
+
```
dcoleman21, line 2: Very nice test

spec/poros/food_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+
+RSpec.describe Food do
+    let!(:hits) { 50000 }
+
```
dcoleman21, line 5: Nice job here

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'rails_helper'
+
+RSpec.describe 'Foods index page', :vcr do
+
+    it 'displays the total number of items returned by the search' do
```
dcoleman21, line 5: I would also add to these tests that once you click `search` you would `expect(current_path).to eq("/foods")`

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    it 'lists the first 10 items returned in the search' do
+        visit root_path
+        fill_in :q, with: "cheddar cheese"
+        click_on "Search"
+
+        expect(page).to have_content("Results on this page: 10")
+    end
+
+    it 'lists the foods attributes' do
+            #ran out of time trying to get test to work, probably needs a div on my index html
```
dcoleman21, line 22: If you are going to be using `within` blocks, then yes, you will need a `div class=` or a `div id=`

app/services/food_service.rb
```diff
@@ -0,0 +1,9 @@
+class FoodService
+    def self.get_foods(food)
+        conn = Faraday.new(url: 'https://api.nal.usda.gov/fdc/v1/') do |faraday|
```
dcoleman21, line 3: I would break this up into at least two methods:

1. self.conn - which is where you will establish your base connection to the api
2. self.get_foods(food) - where you will add the rest of the endpoint uri. 

This will live under a `private`
```
def self.conn
    Faraday.new(url: 'https://api.nal.usda.gov') do |f|
      f.params[:api_key] = ENV['FOOD_DATA_KEY']
    end
end
```
Then your `self.get_foods(food)` might look like this: (this method lives above the `private`)
```
response = conn.get("/fdc/v1/foods/search?query=#{food}&pageSize=10&pageNumber=1")
JSON.parse(response.body, symbolize_names: true)
```

app/poros/food.rb
```diff
@@ -0,0 +1,15 @@
+class Food
+    attr_reader :hits,
+                :gtin_upc,
+                :description,
+                :brand_owner,
+                :ingredients
+
+    def initialize(attributes, total_hits)
+        @hits = total_hits
+        @gtin_upc = attributes[:gtinUpc]
```
dcoleman21, line 10: We can do checks in our PORO for nil values like such:
```
@gtin_upc = attributes[:gtinUpc] ? attributes[:gtinUpc] : 'No response'
```
This is called a ternary

app/facades/food_facade.rb
```diff
@@ -0,0 +1,8 @@
+class FoodFacade 
+    def self.get_foods(search_term)
+        all_foods = FoodService.get_foods(search_term)
+        all_foods[:foods].map do |data|
+            Food.new(data, all_foods[:totalHits])
```
dcoleman21, line 5: Nice job getting all of the things that we want - attributes and totalHits

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,5 @@
+class FoodsController < ApplicationController
+    def index
+        @foods = FoodFacade.get_foods(params[:search])
```
dcoleman21, line 3: you could give yourself access to totalHits and foods in here by doing this:
```
results = FoodFacade.get_foods(params[:search])
@total_hits = results[0]
@foods = results[1]
```
assuming that `get_foods` is giving us an array. 
Then you have access to call `@total_hits` in your view

dcoleman21 (commented):
Overall nice job Nick. I left some comments for you to look over. Let me know if you have any questions.

PR #150, opened 2022-09-09
Thiago Silveira

tig-o:


app/facades/food_search_facade.rb
```diff
@@ -0,0 +1,8 @@
+class FoodSearchFacade
+  def self.get_foods(query)
+    FoodService.get_food_details(query)
+    json[:foods][0..9].map do |food_data|
```
juliet-e, line 4: Consider using the API's query parameter of `pageSize` to limit the food response to 10 instead of having to use Ruby (just a bit cleaner)

spec/facades/food_search_facade_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'rails_helper'
+
+RSpec.describe FoodSearchFacade do
+  it 'returns a list of foods' do
+    foods = FoodSearchFacade.get_foods('sweet potato')
+
+    foods.each do |food|
+      expect(food).to be_a(Food)
```
juliet-e, line 8: You could also add an assertion that this array of Foods has only 10 foods

spec/services/food_service.rb
```diff
@@ -0,0 +1,12 @@
+require 'rails_helper'
+
+RSpec.describe FoodService do
+  it 'can return foods' do
+    user_query = 'sweet potato'
+    result = FoodService.get_food_details(user_query)
+
+    expect(result).to have_key(:foods)
+
+    expect(result[:foods].first).to have_key(:gtinUPC)
```
juliet-e, line 10: This is a great start! I'd expect to see a few more assertions here about the presence of some other keys that will be needed in the PORO

PR #151, opened 2022-09-09
Bryce Simonds

brycesimonds:
Sorry didn't see the slack message to do a PR too, but submitted the form at 11:00am!


app/facades/food_db_facade.rb
```diff
@@ -0,0 +1,14 @@
+class FoodDBFacade
```
juliet-e, line 1: This is a nicely structured facade! To minimize the number of API calls, you could do everything in one method and extract all the data you need from just one call. Consider that Ruby allows you to return more than one object from a method so you could, in your searched_food_results method, return:

`[all_results[0..9], parsed_json[:totalHits]`

In week 5, we'll also have a class on caching which would allow you to keep the organization of 2 different methods (which is nice and clean) and retrieve the totalHits data from the cache rather than making another API call

app/facades/food_db_facade.rb
```diff
@@ -0,0 +1,14 @@
...
+    parsed_json = FoodDBService.get_searched_foods(searched_query)
+    all_results = parsed_json[:foods].map do |food_data|
+      Food.new(food_data)
+    end
+    return all_results[0..9]
+  end
+
+  def self.the_total_hits_for_search(searched_query)
+    parsed_json = FoodDBService.get_top_hits(searched_query)
+    FoodTotalHits.new(parsed_json)
```
juliet-e, line 12: I appreciate that you made a PORO for totalHits! Because it's just a single Integer attribute it's not totally necessary but it doesn't hurt

app/services/food_db_service.rb
```diff
@@ -0,0 +1,19 @@
+class FoodDBService
+  def self.get_searched_foods(search_query)
```
juliet-e, line 2: No need to have 2 separate methods for the different purposes (searched food vs total hits) as they are the same endpoint.

app/services/food_db_service.rb
```diff
@@ -0,0 +1,19 @@
...
+    response = conn.get("https://api.nal.usda.gov/fdc/v1/foods/search?api_key=#{ENV["food_api_key"]}&query=#{search_query}")
+    json = JSON.parse(response.body, symbolize_names: true)
+  end
+
+  def self.get_top_hits(search_query)
+    response = conn.get("https://api.nal.usda.gov/fdc/v1/foods/search?api_key=#{ENV["food_api_key"]}&query=#{search_query}")
```
juliet-e, line 8: Typically, the path here would only need to be `/fdc/v1/foods . . .` since the base URL is defined on line 15. This might not have been working with the connection however because the parameter on line 16 is named X-API-KEY instead of api_key

app/services/food_db_service.rb
```diff
@@ -0,0 +1,19 @@
...
+    json = JSON.parse(response.body, symbolize_names: true)
+  end
+
+private
+
+  def self.conn
+    Faraday.new(url: "https://api.nal.usda.gov") do |faraday|
+      faraday.params['X-API-Key'] = ENV["food_api_key"]
```
juliet-e, line 16: key for this API is `api_key` instead of `X-API-Key` --> the parameter key name will unfortunately be different for each API

spec/facades/food_facade_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'rails_helper'
+
+RSpec.describe 'Food Facade' do
```
juliet-e, line 3: Nicely structured facade tests

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+
+RSpec.describe 'Foods Index Page' do 
+  it 'shows the total number of items returned by the search' do 
+    visit '/'
+
+    fill_in 'q', with: "sweet potatoes"
+
+    click_button 'Search'
+    
+    expect(page).to have_content('Total hits for this search: 49676')
```
juliet-e, line 11: We'd expect to see some more assertions here with the other attributes you display

spec/poros/food_total_hits_poro_spec.rb
```diff
@@ -0,0 +1,76 @@
+require 'rails_helper'
+
+RSpec.describe 'Food Total Hits Poro' do 
+  it 'it exists and has attributes' do 
+    data = {
```
juliet-e, line 5: I appreciate that you shortened the stub a bit. If this were a project with more time, it'd be great to pull this into a fixture but for an assessment these PORO tests look great! For some of these trickier attributes, consider adding some tests/logic in your PORO around the absence  of some of the attributes

spec/services/food_db_service_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+
+RSpec.describe 'Food DB Service' do 
+  it 'retrieves data for search food' do 
+    parsed_json = FoodDBService.get_searched_foods("sweet potatoes")
+
+    expect(parsed_json).to be_a Hash
+    expect(parsed_json[:totalHits]).to be_a Integer
+
+    first_food_object = parsed_json[:foods][0]
```
juliet-e, line 10: Great test assertions

PR #153, opened 2022-09-09
Ben Silverstein

bensjsilverstein:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,6 @@
+class FoodsController < ApplicationController
+  def results
+    @foods = FoodsFacade.food_search(params[:q])
+    @food_matches = FoodsFacade.food_matches(params[:q])
```
juliet-e, line 4: food_matches is a bit of a confusing name -- is this the total number of foods?

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,14 @@
+class FoodsFacade
```
juliet-e, line 1: Nicely structured facade! Consider trying to get both the total hits and the food POROs in just one API call in order to minimize API hits. Remember that Ruby can allow you to return more than one thing from a method. In your food_search method you could return:
`[foods, data[:totalHits]]`

app/services/foods_service.rb
```diff
@@ -0,0 +1,20 @@
+class FoodsService
+
+  def self.food_results(food)
```
juliet-e, line 3: No need to have 2 different methods to make the same API call

app/services/foods_service.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def self.food_search(food)
+    response = connect.get("foods/search?api_key=#{ENV["food_api_key"]}&query=#{food}")
+    json = JSON.parse(response.body, symbolize_names: true)[:foods]
+  end
+
+  private
+
+  def self.connect
+    Faraday.new(url: "https://api.nal.usda.gov/fdc/v1/") do |faraday|
```
juliet-e, line 16: Typically the Faraday connection will just have the *base* URL and no path in it -- i.e. everything up until .gov but not the path pieces after that.

spec/facades/foods_facade_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'rails_helper'
+
+RSpec.describe FoodsFacade do
```
juliet-e, line 3: Nice test! Sometimes these tests can feel ridiculously simple but because we are also unit testing our PORO and service, we don't need to make more assertions here because they are making sure we're getting the right JSON structure back and populating the POROs correctly

spec/features/welcome/search_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'rails_helper'
+
+RSpec.describe "Search" do
```
juliet-e, line 3: Nice robust test assertions here! If this were a project I would worry about how delicate these tests are given that you're using VCR and the data could change (ie. they could add a food result and that number could go up to 49652) but for an assessment this is a great level of robustness

spec/poros/food_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'rails_helper'
+
+RSpec.describe Food do
+  it "exists", :vcr do
+    servce = FoodsService.food_search("sweet potatoes")
```
juliet-e, line 5: No need to make this service call -- by calling the facade, it is going to call the service
However, to make this a pure unit test, it's best practice to create the test data that you pass into the PORO by copying the JSON response sent back from the API and creating a hash here or in a fixture. Checkout the PORO tests in Congress tracker as a reference

spec/rails_helper.rb
```diff
@@ -62,3 +62,9 @@
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
 end
+VCR.configure do |config|
+  config.cassette_library_dir = "spec/fixtures/vcr_cassettes"
+  config.hook_into :webmock
+  config.filter_sensitive_data('movie_db_key') { ENV['e2d1402726608719026ca6424f229bcf'] }
```
juliet-e, line 68: Nice job remembering to filter your API key!

spec/services/foods_service_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'rails_helper'
+
+RSpec.describe FoodsService, :vcr do
+  it "search by keyword" do
+    foods = FoodsService.food_search("sweet potatoes")
+
+    expect(foods).to be_an(Array)
+    expect(foods.first).to be_a(Hash)
+    expect(foods.first).to have_key(:ingredients)
```
juliet-e, line 9: Consider checking for the presence of every key you will need to use in your PORO

PR #155, opened 2022-09-09
Anna Marie Sterling

AMSterling:


spec/services/food_service_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'rails_helper'
+
```
dcoleman21, line 2: Very nice tests

spec/poros/food_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'rails_helper'
+
+RSpec.describe 'Food', type: :poros do
```
dcoleman21, line 3: Nice job

spec/features/index_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'rails_helper'
+
+RSpec.describe 'The food index by search' do
```
dcoleman21, line 3: Really nice job using within blocks to test the lists of foods upon search

spec/facades/food_facade_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'rails_helper'
+
+RSpec.describe 'Food Facade' do
+  it 'returns data for the searched food' do
+    data = FoodFacade.first_10('sweet potatoes')
+
```
dcoleman21, line 6: These are great tests as well. I would just return the data types. Also I would consider using vcr or webmock for stubbing

app/services/food_service.rb
```diff
@@ -0,0 +1,17 @@
+class FoodService
+  def self.total_hits(keyword_string)
+    response = conn.get("foods/search?query=#{keyword_string}&dataType=Branded&pageSize=10&")
```
dcoleman21, line 3: nice job adding the pageSize to 10. Here is another way you could break things down:
```
class FoodService
    to_json('/fdc/v1/foods/search', query: keyword)
  end

  private

  def self.to_json(url, params = {})
    response = conn.get(url) do |f|
      f.params = params
      f.params[:pageSize] = 10
    end
    JSON.parse(response.body, symbolize_names: true)
  end

  def self.conn
    Faraday.new(url: 'https://api.nal.usda.gov') do |f|
      f.params[:api_key] = ENV['FOOD_DATA_KEY']
    end
  end
end
```

app/facades/food_facade.rb
```diff
@@ -0,0 +1,11 @@
+class FoodFacade
+  def self.first_10(keyword_string)
+    @food_items = FoodService.search_10(keyword_string).map do |data|
```
dcoleman21, line 3: How could we maybe make a single method that gives us all the things that we want?


Here is an example:
```
      Food.new(food_result)
    end
  end
```

app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,6 @@
+class FoodsController < ApplicationController
+  def index
```
dcoleman21, line 2: Following the idea of just making a single call with all the info we want (see my comment in your facade). Here is a way we could get all the info with a single call
```
def index
    results = FoodFacade.foods_by_keyword(params[:foods])
    @foods      = results[1]
    @total_hits = results[0]
  end
```

dcoleman21 (commented):
Overall really nice job. I have left comments for you to review. Let me know if you have any questions.

PR #156, opened 2022-09-10
Tyler Caudill PR

dcoleman21:


spec/features/welcome/index_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'rails_helper'
+
+RSpec.describe 'Welcome page' do
+  it 'exists' do
```
dcoleman21, line 4: This is a nice set up to get to `/foods`.

spec/features/foods/index_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+RSpec.describe 'foods page' do
```
dcoleman21, line 3: We will also want to test that we have a list of only 10 foods that contain the ingredient you searched for. And for each of those foods we should see:
```
- The food's GTIN/UPC code
- The food's description
- The food's Brand Owner
- The food's ingredients
```

config/routes.rb
```diff
@@ -1,3 +1,4 @@
 Rails.application.routes.draw do
   root "welcome#index"
+  resources :foods
```
dcoleman21, line 3: I would add to this:
```
resources :food, only: [:index] 
```
Since this is the only action for now.

app/views/foods/index.html.erb
```diff
@@ -0,0 +1 @@
+<h2>Number of items: <%= @foods.count %></h2>
```
dcoleman21, line 1: This is a great start. We will want to iterate over foods here to get our list and all their attributes.

app/services/food_service.rb
```diff
@@ -0,0 +1,17 @@
+class FoodService
+
+  def self.conn
```
dcoleman21, line 3: Nice helper method here

app/services/food_service.rb
```diff
@@ -0,0 +1,17 @@
...
+    Faraday.new(url: 'https://api.nal.usda.gov') do |f|
+      f.params['api_key'] = ENV['food_key']
+    end
+  end
+
+  def self.json_response(response)
```
dcoleman21, line 9: In here I am wondering if you could pass additional params like pageSize to limit 10

app/poros/food.rb
```diff
@@ -0,0 +1,6 @@
+class Food
+  attr_reader :description
+  def initialize(data)
+    @description = description
```
dcoleman21, line 4: I think this would be 
```
@description = data[:description]
```
We would also want upc (GTIN/UPC code), brandOwner, ingredients in here

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,9 @@
+class FoodsFacade
+
+  def self.search_foods
```
dcoleman21, line 3: Does this need an argument? The keyword or query?

app/facades/foods_facade.rb
```diff
@@ -0,0 +1,9 @@
...
+    items = FoodService.search(query)
+    items.each do |item|
```
dcoleman21, line 5: I might use a `.map` here vs `.each` so that you don't need an accumulator.

PR #157, opened 2022-09-12
Justin MidMod

juliet-e:


app/controllers/foods_controller.rb
```diff
@@ -0,0 +1,6 @@
+class FoodsController < ApplicationController
+  def index
```
juliet-e, line 2: Consider how you could get both of these data points with just one API/facade call (hint: in Ruby you can return more than one item from a method). The benefit here is minimizing API hits

app/facades/food_facade.rb
```diff
@@ -0,0 +1,14 @@
+class FoodFacade
+
+  def self.items_returned(search_term)
+    json = FoodService.search(search_term)
+    json[:totalHits]
+  end
+  
+  def self.search(search_term)
+    json = FoodService.search(search_term)
+    json[:foods].first(10).map do |food_data|
```
juliet-e, line 10: Consider taking advantage of API parameters too to limit the # of results to 10

app/facades/food_facade.rb
```diff
@@ -0,0 +1,14 @@
+class FoodFacade
```
juliet-e, line 1: Nicely structured facade

app/poros/food.rb
```diff
@@ -0,0 +1,9 @@
+class Food
+  attr_reader :code, :description, :ingredients, :brand
+  def initialize(data)
+    @code = data[:foodCode]
+    @description = data[:description]
+    @ingredients = data[:foodNutrients].map {|nutrient| nutrient[:nutrientName]}
```
juliet-e, line 6: We were actually expecting you to use the ingredients key in the same hash as these other attributes but this is an interesting approach! Especially when the ingredients key has an empty value

app/services/food_service.rb
```diff
@@ -0,0 +1,11 @@
+class FoodService
+  def self.search(search_term)
+    query = search_term.gsub(' ','%20')
```
juliet-e, line 3: I believe Faraday should be smart enough for this substitution to not be necessary (i.e. you should be able to pass in a query with spaces

app/services/food_service.rb
```diff
@@ -0,0 +1,11 @@
+class FoodService
+  def self.search(search_term)
+    query = search_term.gsub(' ','%20') 
+    response = conn.get("/fdc/v1/foods/search?#{query}")
```
juliet-e, line 4: You have just the search term and not the *name* of the parameter in the url. Instead of `search?query=sweet potatoes` you have `search?sweet potatoes` -- that might be what caused you trouble

